PR #2, opened 2023-07-21
feat(api): add `is_main()`

McPatate:
I need  to check if a repo's revision is the latest on main, adding the code to do that.


McPatate:
Closing for now

PR #8, opened 2023-07-31
feat(api): add lfs metadata

McPatate:
Full lfs object looks like this :
```json
"lfs": {
    "sha256": "5b1da7ebe3bdcc58c91eb9916f7d881791ac2659307f0c383c325361382f2f42",
    "size": 2239866697,
    "pointerSize": 135
}
```

Needed for the repo scanner

### DIFF
--- a/src/api/tokio.rs
+++ b/src/api/tokio.rs
@@ -593,7 +593,6 @@ impl ApiRepo {
     }
 
     /// Get the raw [`reqwest::RequestBuilder`] with the url and method already set
-    /// cf https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html
     /// ```
     /// # use hf_hub::api::tokio::Api;
     /// # tokio_test::block_on(async {


src/api/sync.rs
```diff
@@ -501,12 +507,69 @@ impl ApiRepo{
...
+    /// # tokio_test::block_on(async {
+    /// let api = Api::new().unwrap();
+    /// api.model("gpt2".to_owned())
+    ///     .detailed_info()
+    ///     .query("blobs".to_owned(), "true".to_owned())
+    ///     .send();
+    /// # })
+    /// ```
+    pub fn detailed_info(&self) -> DetailedInfoBuilder {
+        DetailedInfoBuilder::new(self.api.clone(), &self.api.endpoint, &self.repo)
```
Narsil, line 525:
```suggestion
        DetailedInfoBuilder::new(self.api.clone(), &self.repo)
```
endpoint is already in api, no need to duplicate.

Narsil: Actually use `ApiRepo` instead so you need just Self.

src/api/sync.rs
```diff
@@ -501,12 +507,69 @@ impl ApiRepo{
...
+        if let Some(qps) = self.query_params {
+            let qps: Vec<(&str, &str)> = qps
+                .iter()
+                .map(|qp| (qp.0.as_ref(), qp.1.as_ref()))
+                .collect();
+            request = request.query_pairs(qps);
+        }
+        let response = request.call().map_err(Box::new)?;
+
+        Ok(response.into_json()?)
```
Narsil, line 572:
```suggestion
impl DetailedInfoBuilder {
    fn new(api: &ApiRepo) -> Self {
        let url = format!("{}/api/{}", endpoint, repo.api_url());
         let mut request = self.api.client.get(&url);
        Self {
           request
        }
    }
}

impl DetailedInfoBuilder {
    /// Add a query parameter to the request
    pub fn query(mut self, key: String, value: String) -> Self {
        self.request = self.request.query(key, value)
        self
    }

    /// Send the request and get the resulting JSON
    pub fn send(self) -> Result<Value, ApiError> {
        let response = self.request.call().map_err(Box::new)?;
        Ok(response.into_json()?)
```

WDYT ?

src/api/sync.rs
```diff
@@ -84,6 +85,10 @@ pub enum ApiError {
     /// We tried to download chunk too many times
     #[error("Too many retries: {0}")]
     TooManyRetries(Box<ApiError>),
+
+    /// serde_json error
+    #[error("Serde json error: {0}")]
+    SerdeJson(#[from] serde_json::Error),
```
Narsil, line 91: How were we getting away without it before, is it the map(Box?) ?

src/api/tokio.rs
```diff
@@ -591,14 +593,67 @@ impl ApiRepo{
...
+    pub async fn send(self) -> Result<Value, ApiError> {
+        let mut request = self.api.client.get(&self.url);
+        if let Some(qps) = self.query_params {
+            request = request.query(&qps);
+        }
+        let response = request.send().await?;
+        let response = response.error_for_status()?;
 
-        Ok(model_info)
+        Ok(response.json().await?)
```
Narsil, line 656: Same API structure a proposed in Sync for simplicity.

Narsil (commented):
Quite nice.

I propose a different (simpler imho) internals.

Otherwise looks good. Not sure how `cargo fmt` didn't pick up those before.

src/api/sync.rs
Narsil: Maybe even Scratch the builder, and return the `Request` directly.. One less struct and full control over the querying to the user...

Narsil: Then the fn could be rename `info_request` maybe ?

src/api/sync.rs
McPatate: `.into_json` returns an `io::Result<T>` while `serde_json::from_value` returns a `Result<T, serde_json::Error>`

McPatate: And `reqwest::Response::json` returns a `Result<T, reqwest::Error>`

src/api/sync.rs
McPatate: Yep agree, `info_request` sounds good

McPatate: looks better now !

Narsil (approved):
LGTM

PR #12, opened 2023-08-12
Add logging in project based on 'log' library.

evgenyigumnov:
Add stdout logs in tests.

Example stdout logging during tests:

[2023-08-12T08:45:02Z WARN  hf_hub] File with your token is absent by path "C:\\Users\\igumn\\.cache\\huggingface\\token"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ futures = { version = "0.3.28", optional = true}
 thiserror = { version = "1.0.43", optional = true }
 ureq = { version = "2.7.1", optional = true, features = ["native-tls", "json"] }
 native-tls = { version = "0.2.11", optional = true }
+env_logger = "0.10.0"
+log = "0.4.19"
 
 [features]
 default = ["online"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ futures = { version = "0.3.28", optional = true}
 thiserror = { version = "1.0.43", optional = true }
 ureq = { version = "2.7.1", optional = true, features = ["native-tls", "json"] }
 native-tls = { version = "0.2.11", optional = true }
-env_logger = "0.10.0"
 log = "0.4.19"
 
 [features]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,9 +93,6 @@ impl Cache {
         // Remove `"hub"`
         path.pop();
         path.push("token");
-        if !path.exists() {
-            log::warn!("File with your token is absent by path {:?}", path);
-        }
         path
     }
 
@@ -222,7 +219,6 @@ mod tests {
     #[test]
     #[cfg(not(target_os = "windows"))]
     fn token_path() {
-        env_logger::Builder::from_env(env_logger::Env::new().default_filter_or("warn")).init();
         let cache = Cache::default();
         let token_path = cache.token_path().to_str().unwrap().to_string();
         if let Ok(hf_home) = std::env::var("HF_HOME") {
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -34,6 +34,8 @@ futures = { version = "0.3.28", optional = true}
 thiserror = { version = "1.0.43", optional = true }
 ureq = { version = "2.7.1", optional = true, features = ["native-tls", "json"] }
 native-tls = { version = "0.2.11", optional = true }
+env_logger = "0.10.0"
```
Narsil, line 37:
```suggestion
```

This is not supposed to be a library dependency.

src/lib.rs
```diff
@@ -93,6 +93,9 @@ impl Cache {
         // Remove `"hub"`
         path.pop();
         path.push("token");
+        if !path.exists() {
+            log::warn!("File with your token is absent by path {:?}", path);
```
Narsil, line 97: Can we move that log to `api::{sync,tokio}` instead in the error branch while loading from this path ?

This function could be called many times over spamming every caller.

The token path is always the token path, it shouldn't ever warn.

src/lib.rs
```diff
@@ -219,6 +222,7 @@ mod tests {
     #[test]
     #[cfg(not(target_os = "windows"))]
     fn token_path() {
+        env_logger::Builder::from_env(env_logger::Env::new().default_filter_or("warn")).init();
```
Narsil, line 225: Remove those.

src/lib.rs
Narsil, line 97:
```suggestion
            log::info!("Token file not found {:?}", path);
```

`warn` -> `info`. It's perfectly normal to be missing the token file, therefore warn is not appropriate imo.
Also shortened a bit the message.

McPatate:
Can we use tracing instead of env_logger ?

evgenyigumnov:
> Can we use tracing instead of env_logger ?

Yes, this is list of loggers:

simple_logger
simplelog
pretty_env_logger
stderrlog
flexi_logger
call_logger
std-logger
structured-logger
log4rs
fern

You could visit page https://crates.io/crates/log and find all list of loggers

McPatate:
I was talking about the https://lib.rs/crates/tracing crate specifically :)

evgenyigumnov:
> I was talking about the https://lib.rs/crates/tracing crate specifically :)

If you ask my opinion. I think, yes it is possible to use tracing libray for logging like log library.
But what I see:
1. I try to find tracing library using in project code. It is not look like simple logging
2. I never use tracing library in my project. If project manager of "hf-hub" make decision use tracing for logging. Then I could learn tracing library and change my pull request from log to tracing library.

Narsil (approved):
LGTM ! Thanks for the modifications.

Narsil:
I'll go ahead and merge for now, we can revisit later.

IMO this crate should be minimalistic, hence logging (and sparingly) should be OK.
Users can always trace themselves to the actual potential bottlenecks like `get` and `download`.
All other functions should be rather trivial.

McPatate:
For the record, I'm not happy about this 😄 

`tracing` is a standard and used pretty much everywhere for logging. I think we _should_ revisit eventually!

Narsil:
https://crates.io/crates/tracing (~200k/day)
https://crates.io/crates/log (~400k/day)

I don't care that much about it either way, ideally I think we just should just have neither, or behind feature flags.

McPatate:
Ah I didn't think log was as ubiquitous, looks like a sync vs async divide, tokio has a dependency on `tracing` while most sync libraries have `log`.

I also thought log had less features for libraries, but either I missed it at the time, either it caught up.

Narsil:
You live in `async` world ! :)

PR #14, opened 2023-08-16
Add cache repo

Narsil:
Similar to what is being done for `Api` vs `ApiRepo`.

I'm hesitant to even change the internals of `ApiRepo` to make them own a `CacheRepo`.
That would remove a bunch of clone, but potential discrepancy between `self.api.cache.repo(repo)` vs `self.cache_repo`

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -692,6 +692,7 @@ mod tests {
         let tmp = TempDir::new();
         let api = ApiBuilder::new()
             .with_progress(false)
+            .with_token(None)
             .with_cache_dir(tmp.path.clone())
             .build()
             .unwrap();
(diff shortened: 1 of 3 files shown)


McPatate:
It makes sense to me that `Api` owns both `ApiRepo` and `CacheRepo` and that `CacheRepo` is not owned by `ApiRepo`

McPatate:
My bad, forget what I said. Why does `Api` have a cache repo?

Download logic is in ApiRepo, btw there's a discrepancy `download_tempfile` is in `ApiRepo` in `tokio` vs in `Api` for `sync`.

Narsil:
`download_tempfile` is private, therefore consistency doesn't matter. IIRC there's a good reason they don't need the same things.

> Why does Api have a cache repo?

To get the token, and to be able to create a `CacheRepo` :)

I think the way to go to clean this would be to have `Api` be split between the Cache owned (the exposed user level API) and the actualy `InternalApi` which would contain only the necessary things to make actual http calls. Then

`ApiRepo` could own `InternalApi` and `CacheRepo` and we avoid the issue of `ApiRepo` owning both `CacheRepo` and `Api.Cache` (being potentially out of sync)

Issue #18, opened 2023-08-23
Support incremental downloads of large files somehow.


crlf0710:
For larger models (maybe >= 3.0 GB), there's a larger chance that the downloads get aborted by network failures with slower networks. I wonder if it's possible to make the downloads incremental, so restart from the beginning is not needed?

benedikt-schaber:
@crlf0710  If you have not already begun working on this and would not mind, I should be able to create a PR this weekend to introduce the same retry capabilities to sync that the tokio version already has.

@Narsil Regarding resumable downloads, could we create our own version of .incomplete (perhaps in the tmp folder)? Using etag for identification and creating two files for each download, the partial file itself and a meta file containing the chunk size and the successfully downloaded chunks (as a 'checklist' to keep it async). We could then also assume that all successfully downloaded chunks are correct and just add the missing chunks.
We could rechunk the contiguous chunk sequences or require users to reuse the same chunk size.
Does this sound somewhat reasonable/Would such a feature be welcomed? I could then also write the PR for this.

Narsil:
Done in : https://github.com/huggingface/hf-hub/pull/84

PR #22, opened 2023-09-07
feat/reuse blobs

benedikt-schaber:
### Current behavior
When the file/symlink is not found in snapshots the blob is downloaded regardless of whether or not it is already present.
### New behavior
When the file/symlink is not found in snapshots the etag is used to check for the corresponding blob. It is used if already present.
### Potential Issues
- The python library provides users with the option to force the download of the file. If this request is accepted we should likely do the same. I am however unaware of how such options should be introduced in this repository
- Probably not really an issue, existed before this change was introduced: If symlinking generally works on Windows but fails once, we move the file and end up with invalid symlinks (if I am not mistaken?, not 100% sure because it is Windows, but from what I understand of CreateSymbolicLinkW and std::os::windows::fs::symlink_file).

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -485,10 +485,8 @@ impl ApiRepo {
         Ok(pointer_path)
     }
 
-    /// Downloads a remote file (if not already present) into the cache directory
-    /// to be used locally.
-    /// This functions require internet access to verify if new versions of the file
-    /// exist, even if a file is already on disk at location.
+    /// Downloads a remote file into the cache directory to be used locally.
+    /// If the file is already present it will be overwritten.
     /// ```no_run
     /// # use hf_hub::api::sync::Api;
     /// let api = Api::new().unwrap();
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -598,6 +598,42 @@ mod tests {
     }
 
     #[test]
+    #[cfg(target_family = "unix")]
+    fn corruption() {
+        let tmp = TempDir::new();
+        let api = ApiBuilder::new()
+            .with_progress(false)
+            .with_cache_dir(tmp.path.clone())
+            .build()
+            .unwrap();
+
+        let model_id = "julien-c/dummy-unknown".to_string();
+        let downloaded_path = api.model(model_id.clone()).download("config.json").unwrap();
+        assert!(downloaded_path.exists());
+
+        // corrupt the file
+        std::fs::write(&downloaded_path, "corrupted").unwrap();
+        // remove the symlink
+        std::fs::remove_file(&downloaded_path).unwrap();
+        // Get should recreate the symlink, but the corrupted blob should be reused
+        let cached_path = api.model(model_id.clone()).get("config.json").unwrap();
+
+        assert_eq!(cached_path, downloaded_path);
+        assert_eq!(std::fs::read_to_string(&cached_path).unwrap(), "corrupted");
+
+        // Download should overwrite the blob, and thus it should not be corrupted anymore
+        let downloaded_path = api.model(model_id.clone()).download("config.json").unwrap();
+
+        assert_eq!(cached_path, downloaded_path);
+
+        let val = Sha256::digest(std::fs::read(&*downloaded_path).unwrap());
+        assert_eq!(
+            val[..],
+            hex!("b908f2b7227d4d31a2105dfa31095e28d304f9bc938bfaaa57ee2cacf1f62d32")
+        );
+    }
+
+    #[test]
     fn dataset() {
         let tmp = TempDir::new();
         let api = ApiBuilder::new()
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -424,8 +424,65 @@ impl ApiRepo {
         if let Some(path) = self.api.cache.repo(self.repo.clone()).get(filename) {
             Ok(path)
         } else {
-            self.download(filename)
+            self.create_blob_pointer(filename)
         }
+    }
+
+    /// Force download a blob, even if it is already present locally.
+    fn download_blob(
+        &self,
+        filename: &str,
+        url: &str,
+        blob_path: &PathBuf,
+        size: usize,
+    ) -> Result<(), ApiError> {
+        std::fs::create_dir_all(blob_path.parent().unwrap())?;
+
+        let progressbar = if self.api.progress {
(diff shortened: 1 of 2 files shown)


Narsil:
We cannot move things like this.

The name of the method is `download`. It should therefore always `download` something.

`get` will already use the cache when possible and might be what you are looking for.

benedikt-schaber:
@Narsil Get will not, and with current arguments, cannot check if the blob is already present, since we need the etag for it, which we receive from our head request.
Further download states in its doc string:
/// Downloads a remote file (if not already present) into the cache directory
This PR thus adheres to this guideline.

benedikt-schaber:
As a reference the python library huggingface hub does it here:
```
    if os.path.exists(blob_path) and not force_download:
        # we have the blob already, but not the pointer
        if local_dir is not None:  # to local dir
            return _to_local_dir(blob_path, local_dir, relative_filename, use_symlinks=local_dir_use_symlinks)
        else:  # or in snapshot cache
            _create_symlink(blob_path, pointer_path, new_blob=False)
            return pointer_path
```
in hf_hub_download.

Although I realized I misunderstood which get you meant. I could move the check to the Api's get method, we would already have to query the metadata then however, and, if we do not want to requery it in download, it would have to be passed along changing download's signature.

Narsil:
@benedikt-schaber 

Having etag check + download if they do not match we can add (basically free updating).
There's a few consideration I've been thinking about this though:

1. Usually models don't change, forcing network calls on each model load is excruciatingly slow (especially if the hub has any form of hiccup, or if you're anywhere outside of the US). This is far from ideal when most of the time it doesn't do anything.
For instance currently with candle we're able to load entire CPU models in ~10ms, and a single call from europe has 50ms ping. In order to do serverless/lazy local inference it's definitely not great.

2. When models did update you *may* want to instantly update, but you *may* be fine just starting the update on a different thread without blocking the cached load. And just using the new models later once the cache has been fully updated.

Architecturally speaking, copying `huggingface_hub` throughout is not something desirable. There's a lot of necessary legacy in `huggingface_hub` we can avoid here (because the design space is clearer now, and we don't depend on old behavior). For instance `force_download` is imho something we shouldn't have. Different functions should do what you want.

`download` is basically `force_download` 
`get` is basically optional download.

The current goals are:

- Don't mess up under any circumstances mess up existing cache
- Reuse existing cache whenever possible.
- Focus on a minimum subset, not rebuilding the entire thing.

Do you mind sharing a bit more the problem ?

I don't think this is about `incremental downloads` as the original issue states. And maybe thinking a bit about 1. and 2. and how your current setup places itself.

@Wauplin FYI

benedikt-schaber:
@Narsil
Ah, I think I should have made it a bit clearer what this PR does and have provided an example, since there seems to be a misunderstanding:

**What this PR is not:** If the file is already in the cache (snapshot), we do exactly the same thing that we did before the PR. We do NOT check if there is a newer version. Or to be more precise this PR does not have anything to do with detecting updates.

### What this PR is about:
Here is an example, let's say I have my dataset on branch main, my commit is `aaaaaa` and I pull my dataset from there. Sadly a couple of labels for our dataset where wrong, I fix this and create a new commit `bbbbbb`. From 10s of GB only a single file changed, containing the labels. I decide I want to use commit `bbbbbb`now instead of `aaaaaa` in my code.
**currently, without the PR this happens:**
No snapshot exists for `bbbbbb` yet since it is the first time I download the commit. We just check if the files are in snapshot, and if not download them (overriding the exact same data that was already in the blobs folder). This means only 1 file changed, but I now sit idle waiting for 10s of GBs I already had to download.
**with this PR:**
We do the same check for snapshot, which does not exist. But now for every file we also check if it is already in the blobs folder (by etag i.e. its hash). If we realize that we have already downloaded the file (because `aaaaaa` already used it in our case) we just create another symlink to it instead of first overriding it and then creating another symlink to it.

benedikt-schaber:
@Narsil Oh, one more important note (which I realize probably led to the misunderstanding): This is not the PR I wanted to create for #18, I have not yet come around to creating that PR.

Narsil:
Illuminating ! 

Then perfect.
Do you think we could have a smallish test to test the behavior ?

benedikt-schaber:
Yes, I will write one.
Coming back to download/force_download:

This PR currently would create a difference between a download and a force_download.
In case a user's cache gets corrupted we probably should give them the ability to redownload everything.
I.e. even though a file with the correct etag is already there in blobs, overwrite it.

If we want to keep download as force_download, we could introduce a new internal download method, that additionally takes what we need of the metadata as input (e.g. blob path, total length).
We could then do the metadata fetch in get (if necessary) and use the internal download method from there. 
Similarly we could fetch the metadata in download, but then just pass it along without checking to the internal download method.

get would thus no longer use download, but instead the new internal download function and we could retain downloads signature and force_download behaviour without compromising to doing an unecessary and additional head fetch for the metadata when using get.

benedikt-schaber:
### API Changes
I have implemented the functions, only for the sync version for now.
I will add the equivalent to tokio after we got them to a state where they are acceptable.
They are named:
download_blob (currently internal, can expose it): Used by the other functions, downloads/overrides a blob. Now used by download
create_blob_pointer (currently internal, temporary name, looking for feedback here): Similar to download, but only downloads blob (using download_blob) if necessary, otherwise only creates the pointer to it. get now uses this function if it can not find the symlink.

### Testing

I have also created a unit test that tests the behavior of get and download (unix only, since on windows there are no symlinks, so this entire PR becomes rather meaningless).

### Naming

Thoughts on a proper name for create_blob_pointer and why I did not choose any that I could think of:
- get_blob: It not only gets the blob, but also creates the smylink to it/moves it.
- get_blob_pointer: Sounds like it either gets or if not present creates the blob pointer. But it always creates the pointer (more or less) and sometimes downloads the blob

Basically also just get without the snapshot cache lookup first. Could be thus named something alluding to that or could also perhaps be just integrated into the get function itself.

### Potential Issue that existed before this PR, but is related
Considering corruption.
In symlink_or_rename we do nothing if the file is already present at destination. This works fine for Linux. But on windows, I think if download truly is supposed to be force download, then, if we have a corrupted file, would it not already be present at destination? So even though we would have just gone through the effort of downloading a version that is not corrupted, in the end the corrupted version would still be used?

Narsil:
Closing as stale.

PR #24, opened 2023-09-13
Expose token reading.

Narsil:


src/api/tokio.rs
```diff
@@ -96,21 +96,7 @@ impl ApiBuilder {
     /// ```
     pub fn new() -> Self {
```
McPatate, line 97: It could be nice to have the token as an optional argument

McPatate (approved):

src/api/tokio.rs
Narsil: It's already there in the builder actually.

PR #26, opened 2023-09-17
Request through proxy if `http_proxy` environment variable exists

bombless:


Narsil:
This shouldn't be done by this lib.

PR #27, opened 2023-09-18
Adding `ApiBuilder::from_cache` method.

Narsil:
Turns out `dirs::home_dir()` can fail on Android apps (there is no $home_dir
set for them).

This is probably super niche, and I didn't find any good way to detect the location cache folder,
therefore I simply include these new functions to allow creating and API from a functional cache.

This is interesting anyway for anyone wanting to use a custom cache location.

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -113,7 +113,7 @@ impl ApiBuilder {
 
     /// From a given cache
     /// ```
-    /// use hf_hub::api::sync::ApiBuilder;
+    /// use hf_hub::{api::sync::ApiBuilder, Cache};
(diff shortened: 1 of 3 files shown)


McPatate:
I use the [home](https://crates.io/crates/home) crate in llm-ls, as they do in cargo:
```rs
    let home_dir = home::home_dir().ok_or(()).expect("failed to find home dir");
```

Narsil:
This isn't linked to that.

It's just that APPS on mobile (iOS, android) are run as users which do NOT have home directories altogether.
on android at least ther

McPatate (approved):

PR #34, opened 2023-10-17
Add proxy from env.

wh7f:
In mainland China, people usually need to access the international internet through a proxy.
Otherwise, they may encounter connection errors.
For example: error trying to connect: tcp connect error: Connection timed out (os error 110)

So it's necessary to support obtaining proxy configurations from environment variables to facilitate the use by users in mainland China.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ num_cpus = { version = "1.15.0", optional = true}
 tokio = { version = "1.29.1", optional = true, features = ["fs", "macros"] }
 futures = { version = "0.3.28", optional = true}
 thiserror = { version = "1.0.43", optional = true }
-ureq = { version = "2.7.1", optional = true, features = ["native-tls", "json"] }
+ureq = { version = "2.8.0", optional = true, features = ["native-tls", "json", "socks-proxy"] }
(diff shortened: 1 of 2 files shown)


jtsai-quid:
hi @Narsil ,
In our cluster, a proxy is essential for accessing the public network.
We came across this issue while upgrading TGI to 1.1.0, which resulted in the tokenizer being unable to load through the proxy.(see  [here](https://github.com/huggingface/text-generation-inference/issues/1108) and [here](https://github.com/huggingface/tokenizers/issues/1373))
Would you please have a look at this PR?
Many thanks

Narsil (approved):
LGTM.

jtsai-quid:
hi @Narsil,
When is the expected release date for the next version to come with this patch?
Many thanks

dimakuv:
> When is the expected release date for the next version to come with this patch?

I also encountered this issue. Would like to bump the priority of releasing a new version with this patch.

kerthcet:
Kindly ping @Narsil have no idea about hf release schedule, but seems we haven't cut a new one for a long time, maybe worths a new release? We also meet this problem and solved it by enable the `proxy-from-env` feature of ureq, which will enable the proxy underground.

PR #37, opened 2023-12-12
Add means of disabling native-tls

dureuill:
Hello 👋 

Thank you for making the Hugging Face hub, it's super convenient.

This PR adds a means of disabling the `native-tls` dependency, because we're using `rust_tls` and really don't want the native one.

It works by adding a new default feature `native-tls` that adds the existing `native-tls` optional dependencies (both for ureq and as a direct dependency). So by default this PR doesn't change anything.

However, it is now possible to remove the `native-tls` dependencies by adding the `hf-hub` without its default features, and with the `online` feature enabled.

I checked that the resulting cargo tree doesn't have `native-tls` at all in this configuration, while `native-tls` remains when the default features are enabled.

Thank you for reading 🙂


McPatate:
Closing in favor of #53

PR #39, opened 2023-12-20
add size to Siblings

kczimm:
Add the file size to the Siblings struct.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -12,6 +12,8 @@ pub mod sync;
 pub struct Siblings {
     /// The path within the repo.
     pub rfilename: String,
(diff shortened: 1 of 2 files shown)


McPatate (approved):
lgtm, @Narsil ?

Narsil:
This would need to be done to both sync/async parts.

I'd rather make it easer for everyone to get exactly what they want. This is what `info_request()`  should get you.
With `info_request` users can get exactly what they want (including the size or not, or any other information).

I think maybe expanding/showcasing this might be more useful than modifying this for everyone.
We should probably add a link within `info`  too.

Narsil:
Closing as it's done I think.

PR #40, opened 2023-12-21
`derive(Debug)` in most places.

MarcusDunn:
This started as wanting to debug print a `Repo`. It turned into 
- Derive `Debug` on nearly everything
- remove unused `Result` on `build_headers` 
- rename `file1` to `file` in `create_ref`
- two spelling fixes (`buids` -> `builds`, `interacto` -> `interact`)
- change `if` + `panic!` into an `assert_eq!` in `make_relative` (this seems like a bug? I think the assertion is supposed to check both paths are absolute - it currently accepts if both are not absolute paths. I left behavior unchanged)
- added semicolons where the expression was unused.

At one point I had `must_use` strewn around after a `clippy fix` but I decided to remove them to keep consistent with current practices.

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -102,7 +102,6 @@ impl ApiBuilder {
     /// use hf_hub::api::sync::ApiBuilder;
     /// let api = ApiBuilder::new().build().unwrap();
     /// ```
-    #[must_use]
     pub fn new() -> Self {
         let cache = Cache::default();
         Self::from_cache(cache)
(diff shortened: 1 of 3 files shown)


Narsil (approved):
LGTM.

I kind of prefer if !xx.empty() { Some() } else {None} just because it feels easier to have the happy path first in my mind (unless you use the unhappy paths for early returns + deindent.

`make_relative` was taken from somewhere on stackoverflow, ideally I'd replace the panic with a proper error but since it's internal it shouldn't make a big difference.

Thanks for the patch!

Narsil:
Merging this as-is, we can make follow-ups if you want to change the panic into a Result.

PR #42, opened 2024-02-04
Add proxy-from-env feature to ureq crate

liutaon:
The asynchronous version of hf-hub automatically enables proxy support based on environment variables by default, whereas the synchronous version does not. To ensure consistency between the two, it is necessary for the synchronous version to also adopt this behavior.


GitNiko:
Currently, the only way to implement this fix is to make code changes in the local repository before the maintainer merges this PR

Narsil:
This is done on `main`.

PR #44, opened 2024-02-10
Add with endpoint and use rustls-tls instead of native-tls

qooba:
Add `with_endpoint` function which will enable to change default hardcoded endpoint.


Narsil:
Done.

PR #55, opened 2024-04-08
feat: add `tokio-rustls-tls` feature

McPatate:
Adding the capability to use `rustls` when using the `tokio` feature instead of the default tls deps.

Also a bit of cleanup.


Narsil (approved):
LGTM.

PR #57, opened 2024-04-17
Support for resume functionality to avoid the need to re-download lar…

hawklithm:
Support for resume functionality to avoid the need to re-download large files from the beginning in case of network interruptions during the download process.
The compilation switch 'download-with-offset' controls whether to enable the resume functionality, with the default setting being disabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,10 +36,10 @@ ureq = { version = "2.8.0", optional = true, features = [
   "json",
   "socks-proxy",
 ] }
+pget="0.1.2"
 
 [features]
 default = ["default-tls", "tokio", "ureq"]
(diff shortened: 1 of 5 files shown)


hawklithm:
1. support download precess resume
2. parallel download
3. support huggingface mirror site by env setting

Narsil:
Happy to add resumable downloads, however we're not going to add a dependency on an unknown/small dependency.

Ideally there's no dependency at all, this is core functionality of this lib, so we need to be in control of all the tradeoffs (How do we handle different failures, different kind of resume, and detect potential invalid files.)

Narsil:
Done here: https://github.com/huggingface/hf-hub/pull/84

PR #58, opened 2024-04-30
feat: allow endpoint to be configured

raptium:
This update enables the client to be configured with a different endpoint other than `https://huggingface.co`, allowing the use of a mirror with this library.


Narsil:
https://github.com/huggingface/hf-hub/pull/85 Done here (through a different initialization, it shouldn't be done automatically, it should be opted-in by applications to use such environment flags)

PR #50, opened 2024-03-22
Set endpoint value from env var

Adamska1008:
If environment variable `HF_ENDPOINT` exists, use it to build api.
fix #47

PR #59, opened 2024-05-02
Add HF_ENDPOINT environment variable option

pi314ever:
As the title states.

Gives a way to change the HF endpoint similar to the python version.


pi314ever:
Duplicate of #50

pi314ever:
Duplicate of #50

ur4t:
Kindly tip: `env!()` takes effect at compile time and it will not work properly as you expected.

Issue #54, opened 2024-04-03
Token file not found "token" 


harshakhmk:
I am using fastembed-rs package, to create embeddings for text input
While I am running my application as a shared library on Android, it is giving below error

```
hf_hub: Token file not found "token"
ureq:::stream connecting to huggingface.co 


```

```
use hf_hub::api::sync::Api;

let api = Api::new().unwrap();

let repo = api.model("bert-base-uncased".to_string());
let _filename = repo.get("config.json").unwrap();

// filename  is now the local location within hf cache of the config.json file

```

harshakhmk:
Using the above code in my sample rust crate which is built as a shared library and used in Android app, app is crashing

harshakhmk:
@Narsil any leads on this would be highly helpful

polarathene:
Here is the rough flow of what's happening when you call `Api::new()`:

<details>
<summary>Click to view</summary>

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L258-L262

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L99-L108

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/lib.rs#L194-L208

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L117-L129

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/api/sync.rs#L160-L182

For the `cache.token()` call in `from_cache()`:

https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/lib.rs#L41-L68

</details>

---

So what's happening is:
1. A cache directory for HF to use is checked via the ENV `HF_HOME`, otherwise it [defaults to `~/.cache/huggingface/hub`](https://github.com/huggingface/hf-hub/blob/9d6502f5bc2e69061c132f523c76a76dad470477/src/lib.rs#L199-L202) for the cache directory.
[...]

PR #60, opened 2024-05-19
chore: `ApiBuilder` support default without token

polarathene:
When creating an API instance the current methods enforce checking for a local `token` file, even when you intend to provide your own via `with_token()`, thus finding a token is redundant, as is the info log of one missing being emitted.

- `from_cache()` calls `cache.token()` to set the `token` field in the builder.
- You cannot call `with_token()` prior, both `default()` and `new()` methods use the same initialization. Only workaround is to duplicate `from_cache()` and omit the token check.
- The token is later used by the private `build_headers()` method during the final `build()` call.

**This PR changes:**
- The `Default` impl for both sync/async structs to provide the same struct value except for the `token` field, leaving it as `None` by default. `from_cache()` will use the "Struct Update" syntax to spread the remaining default field values (_this communicates more clearly what the method is actually modifying_).
- The order of fields for the `Api` and `ApiBuilder` structs were grouped contextually for better visibility vs interleaved. Handled via a separate commit.

I've tried to minimize affecting anything else, but technically if someone was relying on `ApiBuilder::default()` before instead of `::new()`, the only difference should be that `token` is not potentially set for them. Anyone who is using `::new()` should be unaffected by this change.

---

Related: https://github.com/huggingface/hf-hub/issues/54#issuecomment-2118679106

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -92,7 +92,13 @@ pub struct ApiBuilder {
 
 impl Default for ApiBuilder {
     fn default() -> Self {
-        Self::new()
+        Self {
+            endpoint: "https://huggingface.co".to_string(),
+            url_template: "{endpoint}/{repo_id}/resolve/{revision}/{filename}".to_string(),
+            cache: Cache::default(),
+            progress: true,
+            token: None,
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Narsil:
I do not see the issue with that extra logging.

There is effectively no token in the Cache folder, so the log seems quite correct. It's also `INFO` level since there's nothing wrong with this at all.
Why not adding a log when you are setting the token afterwards ?

That or setting the `token` file directly in your cache folder before passing it to `from_cache` ?.

I would remove the logging statement before doing anything else to be honest if that's the real issue.

PR #61, opened 2024-05-22
Adding native TLS support

### LINKED ISSUE #56: Add support for `native-tls`
It appears that adding `native-tls` as a feature in `ureq` is not sufficient, and that the `native_tls::TlsConnector` must explicitly be used in code.

jamiegl:
Changed `tokio.rs` to allow configuration of TLS in `reqwest` clients. Changed `sync.rs` to allow configuration of TLS in `ureq` agents. Both are configured via the `ApiBuilder`.

The connection type is still defaulted to `rustls`. In the case of `tokio`, this isn't a strictly necessary change because `reqwest` can be configured via features but providing a programmatic interface could be handy.

I still want to make some changes to dependencies but wanted to open this for discussion.
fix #56

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,9 +25,9 @@ log = "0.4.19"
 num_cpus = { version = "1.15.0", optional = true }
 rand = { version = "0.8.5", optional = true }
 reqwest = { version = "0.12.2", optional = true, default-features = false, features = [
-  "json",
+  "json", "native-tls", "rustls-tls"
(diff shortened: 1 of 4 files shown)


Narsil:
This is done using pure features without any API change. 

That allows us to not compile every dependency and let users choose their SSL implementation on their own.
See https://github.com/huggingface/hf-hub?tab=readme-ov-file#ssltls

PR #66, opened 2024-08-09
feat: support socks proxy

### LINKED ISSUE #41: Connection Timeout
In my Windows OS, when I run the example:
```rust
use hf_hub::api::sync::Api;

fn main() {
    let api = Api::new().unwrap();

    let repo = api.model("bert-base-uncased".to_string());
    let _filename = repo.get("config.json").unwrap();
}
```
- It always raise `Connect Timeout` exception. I tried to use proxy, but it didn't work too.
- If using Python `huggingface_hub` package, the exception didn't happen.
image

oovm:
The synchronous interface already supports the socks5 protocol, but the asynchronous interface lacks support.

This pr adds support for the socks5 protocol to `tokio::Api`

Should fix #41

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+.idea/
 /target
 Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ log = "0.4.19"
 num_cpus = { version = "1.15.0", optional = true }
 rand = { version = "0.8.5", optional = true }
 reqwest = { version = "0.12.2", optional = true, default-features = false, features = [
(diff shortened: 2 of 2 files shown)


Narsil:
There is no need for that.

Users can enable it at will without any involvment in this crate. See here : https://github.com/huggingface/hf-hub/pull/81

It's important we do not enable that feature by default, because otherwise:
- We need to add every transient feature of every dependency, which is at least cumbersome.
- We don't want users to be compiling the most complex versions of dependency if they don't need it.

This is to the best of my understanding how feature were designed and created.

PR #70, opened 2024-09-24
feat: download progress callback

newfla:
### Motivation
Hf-hub is used by ML application/framework as entry point for model resources. When integrating hf-hub in libraries or desktop apps, it can be useful to show information about the on-going download process in different ways than a cli progress-bar (i.e. events forwarded to the tauri js side).

### Techinical details
The PR adds the method `ApiBuilder::with_progress_builder` (both sync and tokio modules) that allows to specify a callback everytime the download progress.

The callback are so specified:

-  _sync_: `pub fn with_progress_callback<F>(mut self, callback: F) -> Self
    where
        F: FnMut(ProgressEvent) + 'static`
- _tokio_: `pub fn with_progress_callback<F, Fut>(mut self, callback: F) -> Self
    where
        F: 'static + Send + Sync + Fn(ProgressEvent) -> Fut,
        Fut: Future<Output = ()> + Send + 'static,`

The event is defined as follow:
```
/// The download progress event
#[derive(Debug, Clone, Serialize)]
pub struct ProgressEvent {
    /// The resource to download
    pub url: String,

    /// The progress expressed as a value between 0 and 1
    pub percentage: f32,

    /// Time elapsed since the download as being started
    pub elapsed_time: Duration,

    /// Estimated time to complete the download
    pub remaining_time: Duration,
}
```

Serialize has been derived to simplify the usage in tauri applications.

### Alternatives

- Define `ApiRepo:download(&self, filename: &str, callback: C)`:
   For symmetry with `ApiBuilder::with_progress` I've prefered to implement the method on the ApiBuilder itself.

- Async callback implementation is far away from being ergonomics: a novel approch is under development [`#![feature(async_closure)]`](https://blog.rust-lang.org/inside-rust/2024/08/09/async-closures-call-for-testing.html) but I was not sure to bring it into this PR

### DIFF
--- a/examples/download.rs
+++ b/examples/download.rs
@@ -1,10 +1,9 @@
 #[cfg(not(feature = "ureq"))]
-#[cfg(not(feature="tokio"))]
-fn main() {
-}
+#[cfg(not(feature = "tokio"))]
+fn main() {}
 
 #[cfg(feature = "ureq")]
-#[cfg(not(feature="tokio"))]
+#[cfg(not(feature = "tokio"))]
 fn main() {
     let api = hf_hub::api::sync::Api::new().unwrap();
 
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1,4 +1,6 @@
-use serde::Deserialize;
+use std::time::Duration;
+
+use serde::{Deserialize, Serialize};
 
 /// The asynchronous version of the API
 #[cfg(feature = "tokio")]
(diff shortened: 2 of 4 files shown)


Narsil:
I solved it using a different approach:

`download_with_progress` receives a `Progress` trait.

https://github.com/huggingface/hf-hub/pull/80

Users are in charge of implementing their own trait and can do whatever. There's no need for `Events` simply 3 functions (init when we receive the file size, update when we received a chunk, and finish when the download is done (since there could be some discrepancy in sizes and maybe implementor don't want to maintain their own state for status).

I've implemented things in a few UI to make sure the trait bounds where easy to satisfy.

Having a proper trait makes it much easier to capture extra state (channels for instance) than closures which are imho very hard to make satisfactory capture semantics with some objects.

Would it also solve your issue ?

newfla:
Thanks for #80 @Narsil it's a good solution for sure but I would have preferred a closure-based solution (async-closures have just been stabilized) because I find implementing a trait to be slightly overkill for this use case.
I would also consider providing other informations beyond the temporary size to avoid forcing each application to have to reimplement information such as the percentage or the remaining time.
Infact the version I proposed here provided the same information visible in the indicaf progressbar to facilitate the display of a progressbar on a GUI

Narsil:
> I would also consider providing other informations beyond the temporary size to avoid forcing each application to have to reimplement information such as the percentage or the remaining time.

I disagree. it's trivial to implement, there's no way we incur any overhead if users don't want this information.
There's potentially a lot of overhead added in progress bars if one is not careful. So having a minimal API (the one of `indicatif` pretty much) means less potential for footguns.

PR #71, opened 2024-10-28
Disable rustls's aws-lc-rs feature

flip1995:
When both the `ring` and `aws-lc-rs` features in `rustls` are enabled, the rustls `CryptoProvider` installer from crate features no longer works, as the provider to be used is ambiguous [1].

This is the default configuration and feature set that other dependencies like `reqwest` are using.

[1]: https://docs.rs/rustls/latest/src/rustls/crypto/mod.rs.html#261-282

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,9 @@ rand = { version = "0.8.5", optional = true }
 reqwest = { version = "0.12.2", optional = true, default-features = false, features = [
   "json",
 ] }
-rustls = { version = "0.23.4", optional = true }
+rustls = { version = "0.23.4", optional = true, default-features = false, features = [
+  "ring",
+] }
 serde = { version = "1", features = ["derive"], optional = true }
 serde_json = { version = "1", optional = true }
 thiserror = { version = "2", optional = true }
@@ -43,7 +45,7 @@ default = ["default-tls", "tokio", "ureq"]
 # These features are only relevant when used with the `tokio` feature, but this might change in the future.
 default-tls = []
 native-tls = ["dep:reqwest", "reqwest/default", "dep:native-tls", "ureq/native-tls"]
-rustls-tls = ["dep:rustls", "reqwest/rustls-tls"]
+rustls-tls = ["dep:reqwest", "dep:rustls", "reqwest/rustls-tls"]
 tokio = [
   "dep:futures",
   "dep:indicatif",


Narsil:
Sorry for the delay, but I fail to see the issue.

Is there any way we can reproduce the issue ?

Whatever the bug, I feel like it's a non fixable issue since the issue would arise with the other feature enabled. I think the solution provided in code is the solution, if both are specified, user needs to install the crypto provider manually (or the default). No ?
This is just how features work, they need to be able to be built with both enabled, right?

flip1995:
> I think the solution provided in code is the solution, if both are specified, user needs to install the crypto provider manually

Yes, this is basically the issue. All other crates in the ecosystem have their features set up in a way, that either the aws OR the ring provider is used and never both, depending on which features you set for those crates.

Let me describe the issue in a bit more detail:

`rustls` picks the provider automatically [here](https://github.com/rustls/rustls/blob/f98484bdbd57a57bafdd459db594e21c531f1b4a/rustls/src/crypto/mod.rs#L261), if either `ring` OR `aws-lc-rs` is set as a feature flag. If both features are active, it can't auto-select a provider and the user of any crate that depends on `rustls` (transitively) needs to select the provider. With the `rustls` default features enabled, `aws-lc-rs` is enabled as a provider.

Why is this a problem for hf-hub? Well, it isn't really _your_ problem.  But changing the `rustls` dependency, like I did in this PR, would make it fit better in the ecosystem.

As `hf-hub` with the `rustls-tls` feature
https://github.com/huggingface/hf-hub/blob/edda880aa00c335de89d8837f415893a28bb0b2a/Cargo.toml#L46
chooses to use `reqwest` with the `rustls-tls` feature, `hf-hub` indirectly also chooses `ring` as a provider through the [`reqwest` dependency](https://github.com/seanmonstar/reqwest/blob/ea48da723c17380a7d9cdf791248cd3a196c005a/Cargo.toml#L96). `reqwest`, `hyper`, `tokio`, `ureq`, ... are all setup, so that you can choose `ring` OR `aws-lc-rs` as a crypto provider through feature flags, or use (in the case of `ureq`) `ring` as the default provider w/o the option of picking `aws-lc-rs`.

So when `hf-hub` now enables the default features of `rustls`, while at the same time using `reqwest` as a dependency with `ring` as the provider, `hf-hub` forces all of its users to depend on both `ring` AND `aws-lc-rs`. This installs unnecessary dependencies for downstream users that only want to use `ring`, and forces them to specify a provider.

As the `rustls` only gets installed with the `rustls-tls` feature enabled, I think the best fix is to go the [`ureq` route](https://github.com/algesten/ureq/blob/229b04af7a2768e6037402a2a45a08a301f9aa49/Cargo.toml#L58) and choose `ring` as the provider.

---

To reproduce:

Using `main` branch of this repo:
```bash
$ cargo new hf-hub-dep-test
$ cd hf-hub-dep-test
$ cargo add --git https://github.com/huggingface/hf-hub --branch main --no-default-features --features rustls-tls,tokio,ureq
$ rg "aws-lc-rs" Cargo.lock
42:name = "aws-lc-rs"
1239: "aws-lc-rs",
1273: "aws-lc-rs",
```

Using the branch of this PR:
```bash
$ cargo new hf-hub-dep-test
$ cd hf-hub-dep-test
$ cargo add --git https://github.com/flip1995/hf-hub --branch rustls-features --no-default-features --features rustls-tls,tokio,ureq
$ rg "aws-lc-rs" Cargo.lock
# No output
```

Cargo.toml
```diff
@@ -43,7 +45,7 @@ default = ["default-tls", "tokio", "ureq"]
 # These features are only relevant when used with the `tokio` feature, but this might change in the future.
 default-tls = []
 native-tls = ["dep:reqwest", "reqwest/default", "dep:native-tls", "ureq/native-tls"]
-rustls-tls = ["dep:rustls", "reqwest/rustls-tls"]
+rustls-tls = ["dep:reqwest", "dep:rustls", "reqwest/rustls-tls"]
```
flip1995, line 48: I also added back the `dep:reqwest` to the `rustls-tls` feature, as it didn't make sense to me _not_ to depend on this crate in the `rustls-tls` case. LMK if I missed something.

Narsil:
Can you provide an example where this causes an issue ? I wasn't able to reproduce anything.

`rustls` has their default set on `aws-lc-sys` for good reason. I don't thinking changing around the default makes sense.
The other projects have `ring` as default most likely because of legacy.

tokio has aws-lc as a default: https://github.com/rustls/tokio-rustls/blob/main/Cargo.toml#L20.

You PR comes with the same caveat as the current state. You also force `ring` and cause issue if someone wants to use `aws-lc`.
I think using our dependencies defaults (even if they do not agree) makes it simpler than re-exposing every possible feature, also violating ourselves the promise of additive nature of features.

Now if there's an example of an actual issue not solvable in the current state we can look into it.

flip1995:
> `rustls` has their default set on `aws-lc-sys` for good reason. I don't thinking changing around the default makes sense.
> The other projects have `ring` as default most likely because of legacy.

Correct, `aws-lc-sys` is the default for `ring` and for most of the above mentioned crates. However all crates listed above are providing a way to opt-out of `aws-lc-sys` and into `ring`, if they depend on `rustls` in one way or another. They usually do this by providing a `rustls-tls` feature and disabling their default features, which uses `aws-lc-sys`. The same thing this PR is doing. See for example [`reqwest`](https://docs.rs/crate/reqwest/latest/features), [`hyper-rustls`](https://docs.rs/crate/hyper-rustls/latest/features), and [`tokio-rustls`](https://docs.rs/crate/tokio-rustls/latest/features).

> You PR comes with the same caveat as the current state. You also force `ring` and cause issue if someone wants to use `aws-lc`.

Not quite: The current state forces the `ring` feature through enabling the `reqwest/rustls-tls` feature while disabling  `reqwest`'s default features, when using `hf-hub` with the `rustls-tls` feature. However, by not disabling default features in `rustls`, it _also_ forces `aws-lc-rs`. This PR now _only_ forces `ring` with the `rustls-tls` feature and _only `aws-ls-rs` when used with the `native-tls` feature/default features, bringing it in line with the other crates.

> Now if there's an example of an actual issue not solvable in the current state we can look into it.

`rustls` provides a way to choose the `CryptoProvider` manually with the [`builder_with_provider`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#method.builder_with_provider), if the ["crate features are ambigous"](https://docs.rs/rustls/latest/src/rustls/crypto/mod.rs.html#255-260). So in theory it is solvable in the current state. But I'd argue that the current state is still flawed, which this PR is trying to fix:

You rarely use those builders or setup `rustls` yourself, but rather through one on those dependencies tokio/hyper/tonic/reqwest/... This means, that the following code works and chooses `ring` as the crypto provider:

<details>
<summary>Reproducer code</summary>

```toml
[package]
name = "hf-hub-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
secret-vault = { version = "^1", default-features = false, features = ["gcp-secretmanager"] }
tokio = { version = "1.42.0", features = ["rt", "macros"] }
```

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    secret_vault::gcp::GcpSecretManagerSource::new("my-project").await?;

    Ok(())
}
```
</details>

But, if then adding `hf-hub` as a dependency, choosing `rustls-tls` over `native-tls` as a feature, which for all other crates in the ecosystem means use `ring`, not `aws-lc-rs` as the provider, the same code as above stops working.

```bash
cargo add hf-hub --features rustls-tls,tokio,ureq --no-default-features
```


<details>
<summary>Example backtrace</summary>

```bash
$ cargo run
thread 'main' panicked at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.20/src/crypto/mod.rs:249:14:
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: core::panicking::panic_display
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:264:5
   3: core::option::expect_failed
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:2025:5
   4: core::option::Option<T>::expect
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:928:21
   5: rustls::crypto::CryptoProvider::get_default_or_install_from_crate_features
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.20/src/crypto/mod.rs:248:24
   6: rustls::client::client_conn::ClientConfig::builder_with_protocol_versions
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.20/src/client/client_conn.rs:300:13
   7: rustls::client::client_conn::ClientConfig::builder
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rustls-0.23.20/src/client/client_conn.rs:277:9
   8: tonic::transport::channel::service::tls::TlsConnector::new
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tonic-0.12.3/src/transport/channel/service/tls.rs:36:23
   9: tonic::transport::channel::tls::ClientTlsConfig::into_tls_connector
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tonic-0.12.3/src/transport/channel/tls.rs:120:9
  10: tonic::transport::channel::endpoint::Endpoint::tls_config
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tonic-0.12.3/src/transport/channel/endpoint.rs:251:17
  11: gcloud_sdk::api_client::GoogleEnvironment::init_google_services_channel::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gcloud-sdk-0.26.1/src/api_client.rs:208:13
  12: gcloud_sdk::api_client::GoogleApiClient<B,C>::with_token_source::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gcloud-sdk-0.26.1/src/api_client.rs:51:87
  13: gcloud_sdk::api_client::GoogleApiClient<gcloud_sdk::api_client::GoogleApiClientBuilderFunction<C>,C>::from_function_with_token_source::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gcloud-sdk-0.26.1/src/api_client.rs:141:10
  14: gcloud_sdk::api_client::GoogleApiClient<gcloud_sdk::api_client::GoogleApiClientBuilderFunction<C>,C>::from_function_with_scopes::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gcloud-sdk-0.26.1/src/api_client.rs:121:10
  15: gcloud_sdk::api_client::GoogleApiClient<gcloud_sdk::api_client::GoogleApiClientBuilderFunction<C>,C>::from_function::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gcloud-sdk-0.26.1/src/api_client.rs:105:10
  16: secret_vault::gcp::gcp_secret_manager_source::GcpSecretManagerSource::with_options::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/secret-vault-1.15.0/src/gcp/gcp_secret_manager_source.rs:44:14
  17: secret_vault::gcp::gcp_secret_manager_source::GcpSecretManagerSource::new::{{closure}}
             at /home/pkrones/.cargo/registry/src/index.crates.io-6f17d22bba15001f/secret-vault-1.15.0/src/gcp/gcp_secret_manager_source.rs:34:10
  18: hf_hub_repro::main::{{closure}}
             at ./src/main.rs:3:66
```
</details>

Narsil:
Thanks for the reproducer, much easier to get, I'm not sure why the test don't trigger the bug since they also enable both features and should be triggerring SSL installation. Do you have ideas why we can't see the same behavior ?
It could be something because test compilation has different flags ?

Again, I stand still that your solution also makes it harder since I can't really use `aws-lc-rs` without implementing the crypto provider.

But I don't care one way or the other which crypto backend is used, ultimately it is up to users to care about this. 

Here is another attempt which fixes your issue without forcing any implementation :https://github.com/huggingface/hf-hub/pull/88
So whenever this passes: https://github.com/seanmonstar/reqwest/issues/2423 then at least we will seamlessly change too.

Would this work for your use case ? This remove the rustls dependency altogether from this crate and it just transitively let other crates check. I'm not sure why we added `rustls` in the first place.

flip1995:
> Again, I stand still that your solution also makes it harder since I can't really use `aws-lc-rs` without implementing the crypto provider.

My solution brings it in line with all the other crates in the ecosystem. `aws-lc-rs` can easily be chosen by adding the `rustls` dependency on user side, enabling the `aws-lc-rs` feature and setting the provider. But currently `hf-hub` _makes_ this decision for the user.



> Would this work for your use case ?

Yes, removing the direct dep on `rustls` would solve the problem as well. If that works, I think it is the better solution!

Narsil:
Closing in favor of https://github.com/huggingface/hf-hub/pull/88 which also removes the implicit `aws-lc-rs` dep, letting either `reqwest` or `ureq` choose the default.

flip1995:
Thanks for fixing this! Is there a release process for `hf-hub`? Not sure if you think this change is worth to do a release for, but I'd like to avoid `git` dependencies in my `Cargo.toml` file :)

PR #77, opened 2024-12-16
add native-tls

### LINKED ISSUE #56: Add support for `native-tls`

neo773:
This PR added `native-tls` support  unlocking use cases where the user maybe behind a corporate firewall.

Fixes #56

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,10 +35,12 @@ tokio = { version = "1.29.1", optional = true, features = ["fs", "macros"] }
 ureq = { version = "2.8.0", optional = true, features = [
   "json",
   "socks-proxy",
+  "native-tls"
 ] }
+native-tls = { version = "0.2", optional = true }
 
 [features]
-default = ["default-tls", "tokio", "ureq"]
+default = ["default-tls", "tokio", "ureq", "native-tls"]
 # These features are only relevant when used with the `tokio` feature, but this might change in the future.
 default-tls = ["dep:reqwest", "reqwest/default"]
 rustls-tls = ["dep:reqwest", "dep:rustls", "reqwest/rustls-tls"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -35,10 +35,12 @@ tokio = { version = "1.29.1", optional = true, features = ["fs", "macros"] }
 ureq = { version = "2.8.0", optional = true, features = [
   "json",
   "socks-proxy",
+  "native-tls"
 ] }
+native-tls = { version = "0.2", optional = true }
 
 [features]
-default = ["default-tls", "tokio", "ureq"]
+default = ["default-tls", "tokio", "ureq", "native-tls"]
```
Narsil, line 43: I'm confused by this ?

 Why adding another tls feature ?
 
 `default-tls` is defined in reqwest. For now it means native-tls.
 The idea behind it is that the default could change (breaking semver) at some point. I think we should keep that semantic.
 
 So `native-tls` could exist, but it should be exclusive with `default-tls` and would mean exactly : get me native-tls.

src/api/sync.rs
```diff
@@ -163,21 +166,45 @@ impl ApiBuilder {
     pub fn build(self) -> Result<Api, ApiError> {
         let headers = self.build_headers();
 
+        // Create the TLS connector when native-tls feature is enabled
+        #[cfg(feature = "native-tls")]
+        let agent = {
```
Narsil, line 171: I think we should factor out this so that there's only 1 compile time flag (just so that it's easier to reason about things).

I can take care of it if you want.

neo773: > I can take care of it if you want.

Sure, that'd be great Rust isn't my main stack

Narsil: I've done that here: https://github.com/huggingface/hf-hub/commit/b79b16fadf3540b3f920080c14fcf6d5980ffdd1

Happy to put you as a co-author in some related PR to give back credit. Thanks for your contribution.

neo773: Sure, thank you.

Narsil:
This is already merged in `main` in a different form.

Closing for now.

PR #84, opened 2024-12-27
Resumable downloads.

Narsil:
Notes on the strategy here.

Sync:
- Relatively straigforward, we're not multiplexing, we write in order, therefore we can simply pick up whereever we left off.

Tokio:
Since we're multiplexing, writes are not happening in order. We could:
- Store every written interval and skip those, but that involves creating somewhat complex datastructures.
- Instead, since writes are *mostly* ordered, we simply commit whatever was fully written as a single `u64` at the very end of the partial file. When resuming, we read that single int, and resume from there. Anything fishy and we drop the resumability.

This feels simpler that the more complex version, seems to work quite well in practice (only tested on stable networks right now).
The probability for storing invalid files seems low (since commit happens at a single location both in code and in file).

### DIFF
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -29,7 +29,7 @@ type HeaderMap = HashMap<&'static str, String>;
 type HeaderName = &'static str;
 
 /// Specific name for the sync part of the resumable file
-const EXTENTION: &'static str = ".part";
+const EXTENTION: &str = ".part";
 
 struct Wrapper<'a, P: Progress, R: Read> {
     progress: &'a mut P,
@@ -429,7 +429,7 @@ impl Api {
         tmp_path: PathBuf,
         filename: &str,
     ) -> Result<PathBuf, ApiError> {
-        progress.init(size as usize, filename);
+        progress.init(size, filename);
         let filepath = tmp_path;
 
         // Create the file and set everything properly
--- a/src/api/sync.rs
+++ b/src/api/sync.rs
@@ -707,6 +707,7 @@ mod tests {
     use rand::{distributions::Alphanumeric, Rng};
     use serde_json::{json, Value};
     use sha2::{Digest, Sha256};
+    use std::io::{Seek, SeekFrom, Write};
 
     struct TempDir {
         path: PathBuf,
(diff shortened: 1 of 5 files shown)


src/api/sync.rs
```diff
@@ -419,15 +426,26 @@ impl Api {
...
     ) -> Result<PathBuf, ApiError> {
         progress.init(size, filename);
-        let filepath = self.cache.temp_path();
+        let filepath = tmp_path;
 
         // Create the file and set everything properly
-        let mut file = std::fs::File::create(&filepath)?;
 
-        let mut res = self.download_from(url, 0u64, size, &mut file, filename, &mut progress);
+        let mut file = match std::fs::OpenOptions::new().append(true).open(&filepath) {
```
McPatate, line 437:
```suggestion
        let mut part_file = match std::fs::OpenOptions::new().append(true).open(&filepath) {
```

src/api/tokio.rs
```diff
@@ -492,32 +502,61 @@ impl ApiRepo {
...
-            .set_len(length as u64)
-            .await?;
+        const N_BYTES: usize = size_of::<u64>();
+        let start = match tokio::fs::OpenOptions::new()
+            .read(true)
+            .open(&filename)
+            .await
+        {
+            Ok(mut f) => {
+                let len = f.metadata().await.unwrap().len();
```
McPatate, line 519: `unwrap` expected here?

src/api/tokio.rs
```diff
@@ -492,32 +502,61 @@ impl ApiRepo {
...
+                        let committed = u64::from_le_bytes(buf);
+                        committed as usize
+                    } else {
+                        0
+                    }
+                } else {
+                    0
+                }
+            }
+            Err(_err) => {
```
McPatate, line 534:
```suggestion
            Err(_) => {
```

McPatate (commented):
1st part of my review, trying to figure out how things work in `tokio`

McPatate (approved):
lgtm, understood!

src/api/sync.rs
```diff
@@ -739,6 +760,85 @@ mod tests {
...
+        );
+
+        let blob = std::fs::canonicalize(&downloaded_path).unwrap();
+        let file = std::fs::OpenOptions::new().write(true).open(&blob).unwrap();
+        let size = file.metadata().unwrap().len();
+        let truncate: f32 = rand::random();
+        let new_size = (size as f32 * truncate) as u64;
+        file.set_len(new_size).unwrap();
+        let mut blob_part = blob.clone();
+        blob_part.set_extension(".part");
```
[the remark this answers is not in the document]
Narsil: It's done on purpose to have it hardcoded in the test. I don't want my test to change when library code changes.

src/api/tokio.rs
Narsil: Fixed.

Issue #117, opened 2025-07-16
make Hf metadata part of the public API


michaelfeil:
It would be awesome to access the metadata of HfAPI. 

async fn metadata is currently private.

Narsil:
Sorry, I don't think we want to do that.

PR #118, opened 2025-07-16
make metadata part of public api

michaelfeil:
Closes https://github.com/huggingface/hf-hub/issues/117


Narsil:
I'd rather keep that internal. This struct is highly dependent on the hub's internals. Exposing it means breaking whenever we need to modify it, which currently wouldn't break anything, like if we want to start becoming mirror aware or stuff like that.

If you need it, just copy the struct yourself. Normally with the raw client you can issue the same calls as us.

PR #126, opened 2025-09-23
Update indicatif to current version in Cargo.toml

gordonmessmer:
The indicatif minor version was bumped due to a change in its console dep, but shouldn't have any other breaking changes.

Updating this dep past 0.17 will make it easier to ship this crate in Fedora.


Narsil (approved):
LGTM

PR #136, opened 2026-01-18
Expose `metadata` and `pointer_path` methods

danieldk:
I have used/needed these useful methods in [safetensors-browser](https://github.com/danieldk/safetensors-browser). Making them visible in the official crate makes it possible to upload safetensors-browser to crates.io again.

### DIFF
--- a/src/api/tokio.rs
+++ b/src/api/tokio.rs
@@ -1354,6 +1354,7 @@ mod tests {
             json!({
                 "_id": "621ffdc136468d709f17ddb4",
                 "author": "mcpotato",
+                "config": {},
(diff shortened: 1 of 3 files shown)


src/api/tokio.rs
```diff
@@ -1432,7 +1452,7 @@ mod tests {
             .build()
             .expect("failed to build API");
 
-        let repo = api.model("google-bert/bert-base-uncased".to_string());
+        let repo = api.model("danieldk/private-test".to_string());
```
danieldk, line 1455: `google-bert/bert-base-uncased` seems to be public now, so I just made a private repo.

McPatate (approved):
lgtm!

PR #138, opened 2026-01-28
Fix CJK filename crash bug in message formatting

darkautism:
If repo contains CJK filename, hf-hub will crash.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -43,7 +43,12 @@ impl Progress for ProgressBar {
             );
         let maxlength = 30;
         let message = if filename.len() > maxlength {
-            format!("..{}", &filename[filename.len() - maxlength..])
+            // Fix CJK filename crash bug
+            let start = filename.char_indices()
+            .map(|(i, _)| i)
+            .find(|&i| filename.len() - i <= maxlength)
+            .unwrap_or(filename.len());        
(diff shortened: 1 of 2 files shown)


src/api/mod.rs
```diff
@@ -43,7 +43,12 @@ impl Progress for ProgressBar {
             );
         let maxlength = 30;
         let message = if filename.len() > maxlength {
-            format!("..{}", &filename[filename.len() - maxlength..])
+            // Fix CJK filename crash bug
```
Narsil: Something like that seems simpler imho:

```rust
fn elide_start_ref(input: &str, max_chars: usize) -> &str {
    let char_count = input.chars().count();

    if char_count <= max_chars {
        return input;
    }

    // Calculate how many characters to skip from the start
    let skip_count = char_count - max_chars;

    // Find the byte index where the 'skip_count'-th character starts
    // char_indices() returns (byte_index, char)
    if let Some((byte_idx, _)) = input.char_indices().nth(skip_count) {
        &input[byte_idx..]
    } else {
        ""
    }
}
```

darkautism: I change elide_start_ref to do this.

Narsil: I said something like, not copy paste. Please exercise judgement calls about how this code can look better.

PR #145, opened 2026-03-18
Add xet upload/download support to hf-hub

assafvayner:
## Summary

- Implement xet-aware downloads and full upload API for hf-hub's tokio async API
- **Xet-aware downloads**: detect xet files via `X-Xet-Hash` HEAD response header, download via `XetSession` instead of HTTP
- **Full upload pipeline**: `upload_file`, `upload_folder`, `create_commit` with server-negotiated routing — small files inline via HTTP, large files via xet
- **New modules**: `commit.rs` (CommitOperation types, preupload, create_commit), `lfs.rs` (UploadInfo, LFS batch), `xet.rs` (session lifecycle, token refresh, download/upload helpers)
- Auth follows the latest JSON-based xet token spec with `TokenRefresher` for autonomous refresh
- Shared `XetSession` on `Api` with double-checked locking via `Arc<RwLock>`
- `Metadata.size` changed from `usize` to `u64` for >4GB file support
- All xet code gated behind optional `xet` feature flag (off by default)

## Test plan

- [x] `cargo check` passes with default features
- [x] `cargo check --features xet` passes
- [x] All 22 existing tests pass (4 pre-existing header version string failures unrelated to this PR)
- [ ] Integration test: download a known xet-stored file from a public repo
- [ ] Integration test: upload a file via `upload_file`, download it back, verify round-trip
- [ ] Integration test: `create_commit` with mixed operations (add + delete + copy)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
+++ b/docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
@@ -398,7 +398,7 @@ pub struct Api {
     max_retries: usize,
     progress: bool,
     #[cfg(feature = "xet")]
-    xet_session: Option<XetSession>,  // None until first xet op; cheap to clone
+    xet_session: Arc<RwLock<Option<XetSession>>>,  // written once, read many
 }
 
 pub struct ApiRepo {
@@ -407,25 +407,50 @@ pub struct ApiRepo {
 }
 ```
 
-`ApiRepo` accesses the session via a helper that initializes if needed and stores on `Api`:
+`Api` exposes a `xet_session()` method using double-checked locking via `RwLock`:
 
 ```rust
 #[cfg(feature = "xet")]
-async fn ensure_xet_session(&mut self) -> Result<&XetSession, ApiError> {
-    if self.api.xet_session.is_none() {
-        let session = xet::create_session(
-            &self.api.client,
-            &self.api.endpoint,
-            /* repo_type, repo_id, revision from self.repo */
-            &self.api.headers,
-        ).await?;
-        self.api.xet_session = Some(session);
+impl Api {
+    /// Returns a cloned XetSession, initializing it on first call.
+    /// Uses double-checked locking: acquires read lock first, falls back
+    /// to write lock only if session is uninitialized.
+    pub(crate) async fn xet_session(
+        &self,
+        repo_type: &str,
+        repo_id: &str,
+        revision: &str,
+    ) -> Result<XetSession, ApiError> {
+        // Fast path: read lock, return clone if already initialized
+        {
+            let guard = self.xet_session.read().await;
+            if let Some(ref session) = *guard {
+                return Ok(session.clone());
+            }
+        }
+        // Slow path: write lock, double-check, then create
+        {
+            let mut guard = self.xet_session.write().await;
+            // Another task may have initialized while we waited for the write lock
+            if let Some(ref session) = *guard {
+                return Ok(session.clone());
+            }
+            let session = xet::create_session(
+                &self.client,
+                &self.endpoint,
+                repo_type,
+                repo_id,
+                revision,
+                /* headers */
+            ).await?;
+            *guard = Some(session.clone());
+            Ok(session)
+        }
     }
-    Ok(self.api.xet_session.as_ref().unwrap())
 }
 ```
 
-Since `Api` is `Clone` and `XetSession` is cheaply cloneable, cloning an `Api` after session initialization carries the session to all clones.
+`ApiRepo` calls `self.api.xet_session(...)` to get a session clone. Since `XetSession` is cheaply cloneable (Arc-based internally), the returned clone shares all state with the original.
 
 ### Download branching in `download_with_progress`
 
@@ -443,8 +468,8 @@ pub async fn download_with_progress<P: Progress + Clone + Send + Sync + 'static>
 
     #[cfg(feature = "xet")]
     if let Some(ref xet_data) = metadata.xet_file_data {
-        let session = self.ensure_xet_session().await?;
-        xet::xet_download(session, xet_data, metadata.size, &blob_path).await?;
+        let session = self.api.xet_session(/* repo_type, repo_id, revision */).await?;
+        xet::xet_download(&session, xet_data, metadata.size, &blob_path).await?;
     } else {
         self.http_download(&url, &metadata, &blob_path, progress).await?;
     }
@@ -536,7 +561,7 @@ create_commit(params)
   │
   ├── 4. For lfs_files:
   │     ├── post_lfs_batch_info() → server confirms xet transfer
-  │     ├── self.ensure_xet_session() → get/init shared session
+  │     ├── self.api.xet_session() → get/init shared session
   │     └── xet_upload(&session, lfs_files) → returns xet hashes per file
   │
   └── 5. create_commit_request()
--- a/docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
+++ b/docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
@@ -26,16 +26,12 @@ This design adds two capabilities to hf-hub's tokio API:
 [dependencies]
 hf-xet = { git = "https://github.com/huggingface/xet-core", rev = "c0f798061658259275557df6e124f64973fdcf85", optional = true }
 xet-client = { git = "https://github.com/huggingface/xet-core", rev = "c0f798061658259275557df6e124f64973fdcf85", optional = true }
-typed-builder = { version = "0.20", optional = true }
-
 [features]
 xet = ["dep:hf-xet", "dep:xet-client", "tokio"]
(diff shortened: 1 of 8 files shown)


docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
+# Xet Upload & Download Support for hf-hub
+
+**Date:** 2026-03-18
+**Status:** Approved
+**Scope:** Add xet-aware file downloads and full upload support (with xet for large files) to the hf-hub Rust crate's tokio async API.
```
McPatate, line 5: Do we ignore sync for now?

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+
+```toml
+[dependencies]
+hf-xet = { git = "https://github.com/huggingface/xet-core", rev = "c0f798061658259275557df6e124f64973fdcf85", optional = true }
+xet-client = { git = "https://github.com/huggingface/xet-core", rev = "c0f798061658259275557df6e124f64973fdcf85", optional = true }
+typed-builder = { version = "0.20", optional = true }
+
+[features]
+xet = ["dep:hf-xet", "dep:xet-client", "tokio"]
+tokio = [..., "dep:typed-builder"]
```
McPatate: Don't we want to enable by default?

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+    pub objects: Vec<LfsObject>,
+}
+
+pub(crate) struct LfsObject {
+    pub oid: String,
+    pub size: u64,
+}
+```
+
+We always offer `["basic", "multipart", "xet"]` in the transfers list. If the server doesn't choose `"xet"`, we return `ApiError::UnsupportedTransfer`. Actual LFS upload (basic/multipart) is not implemented in this iteration — see Known Limitations.
```
McPatate, line 224: We could have a hf transfer like fallback for multi part, I think I had started exploring a while ago but since there wasn't much ask for it never completed it.

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+The `XetSession` is long-lived and reused across all operations. Each upload creates an `UploadCommit` and each download creates a `DownloadGroup` from the same session — these are the short-lived, per-operation objects.
+
+## Download Flow Changes (`tokio.rs`)
+
+### Extended FileMetadata
+
+Renamed from `Metadata` to `FileMetadata` for clarity. The `size` field is changed to `u64` to correctly handle files >4GB on 32-bit platforms.
+
+```rust
+pub struct FileMetadata {
```
McPatate: This is ok imo but breaking and will require a major version bump.

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+```
+
+Existing HTTP download logic is extracted into a private `http_download` helper method.
+
+## Upload Flow (`tokio.rs` public API)
+
+### Params structs (typed-builder)
+
+```rust
+#[derive(TypedBuilder)]
```
McPatate: Not super enthused about the extra dep here, but no strong opinion

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+
+### Feature flag testing
+- `cargo test` (default features, no xet) — existing download tests pass, no xet code compiled
+- `cargo test --features xet` — full suite including xet paths
+
+## Known Limitations
+
+1. **No LFS upload fallback**: If the LFS batch server chooses `basic` or `multipart` instead of `xet`, the upload fails with `UnsupportedTransfer`. This means repos without xet enabled cannot use the upload API for large files. Basic LFS upload can be added in a future iteration.
+
+2. **Path-only uploads**: `CommitOperationAdd` only accepts `PathBuf` (file on disk). In-memory `Vec<u8>` / streaming uploads are not supported in this iteration. The type can be extended to an enum (`Source::Path(PathBuf)` / `Source::Bytes(Vec<u8>)`) later.
```
McPatate, line 659: The from memory could definitely be interesting!

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
```diff
@@ -0,0 +1,678 @@
...
+
+3. **Xet download progress**: The xet download path does not integrate with hf-hub's `Progress` trait. `XetSession` has its own internal progress tracking, but it is not surfaced through the existing `download_with_progress` callback. This can be bridged in a future iteration by implementing `TrackingProgressUpdater` as an adapter.
```
McPatate, line 661: ➕ on that as well!

McPatate (commented):
Nice! Not sure we want to merge the doc, but interesting way of discussing a feature ahah

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
assafvayner: oops, I thought claude introduced this struct, I will revert this rename.

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
assafvayner: good question, xet-core will pull a lot of deps/longer compile times.
I think it should be off by default for usage that doesn't require downloading/uploading contents. do you think that's a concern too?

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
assafvayner: Good point, I think we should ignore for this PR and address later

docs/superpowers/specs/2026-03-18-xet-upload-download-design.md
assafvayner: removing it for now

assafvayner:
### Code review

Found 1 issue:

1. **XetSession token type collision: Read session reused for Write operations.** The `Api` struct stores a single `Arc<RwLock<Option<XetSession>>>`. The double-checked locking in `xet_session_with_token_type` returns the cached session on the fast path regardless of the requested `token_type`. If a download (Read token) initializes the session first, a subsequent upload (Write token) will silently reuse the Read-token session and fail with CAS auth errors. The `token_type` parameter is only consulted during first initialization — after that it is ignored. The session needs to be keyed by token type, or separate Read/Write sessions should be stored.

https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/tokio.rs#L693-L722

Additionally, four issues scored just below the reporting threshold but are worth noting:

- `UploadInfo::from_path` reads the entire file into memory via `tokio::fs::read(path)` to compute SHA256. For multi-GB model files (the primary xet/LFS use case), this will OOM. SHA256 should be computed incrementally via streaming. ([lfs.rs L18](https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/lfs.rs#L17-L19))

- Both `xet_session()` and `xet_session_with_token_type()` pass raw `self.repo.revision()` to `fetch_xet_token`, which embeds it unencoded in a URL path. The `url_revision()` method exists to percent-encode slashes for revisions like `refs/pr/2`. Other API calls in the same PR correctly use `url_revision()`. ([tokio.rs L1019](https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/tokio.rs#L1017-L1021), [xet.rs L51](https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/xet.rs#L50-L53))

- Regular (non-LFS) files are read twice during upload: once in `UploadInfo::from_path` for SHA256, and again in `create_commit_request` for base64 encoding. ([commit.rs L203](https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/commit.rs#L202-L204))

- `walkdir` uses `path.is_dir()` which follows symlinks, potentially causing infinite recursion with circular symlinks during `upload_folder`. ([tokio.rs L1324](https://github.com/huggingface/hf-hub/blob/d5ffe82e46dd48db352ca848077353a671b2b5d2/src/api/tokio.rs#L1323-L1326))

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #146, opened 2026-04-01
Add Rust-style cache and search APIs with HF parity

i386:
## Summary
- add a Rust-style cache API under `hf_hub::cache`, with `CacheInfo`, `CachedRepo`, `CachedRevision`, `CachedFile`, and `CacheDeletionPlan`
- add repo file-read APIs on cached repos while dropping uncached fallback reads so the API stays explicit about cache-only behavior
- add a Rust-style Hub search API via `api.search(RepoType::...)`, with builder methods like `with_query`, `with_author`, `with_filter`, `with_limit`, and typed `RepoSummary` results
- align cache path behavior with `huggingface_hub`, including env precedence for `HF_HUB_CACHE`, `HUGGINGFACE_HUB_CACHE`, `HF_HOME`, and `XDG_CACHE_HOME`
- honor `HF_TOKEN_PATH` and preserve unresolved env-var references during path expansion instead of silently dropping them
- normalize explicit `CacheInfo::scan_dir(Some(path))` inputs by expanding `~`, resolving relative paths, and canonicalizing when possible
- validate cache repo folder names against the canonical on-disk prefixes only: `models--`, `datasets--`, and `spaces--`
- refresh tests so version assertions are stable, search has smoke coverage in both backends, and gated redirect tests skip cleanly when the local token lacks model access

## Why
This branch started from a narrow cache-introspection need, but the useful shape for `hf-hub` is broader:
- keep the public API Rust-like instead of mirroring Python names directly
- still align overlapping behavior with the official `huggingface_hub` client where that affects correctness or user expectations

That leads to three complementary surfaces:

```rust
let cache = hf_hub::cache::CacheInfo::scan_dir(None)?;
```

```rust
let results = api
    .search(RepoType::Model)
    .with_query("gpt2")
    .with_limit(5)
    .run()?;
```

```rust
let bytes = cache
    .repo(Repo::model("openai-community/gpt2".to_string()))
    .read_file("config.json")?;
```

Python exposes search through `list_models`, `list_datasets`, and `list_spaces` with a `search=` parameter. In Rust, a small builder reads more naturally and fits the rest of the crate better.

The parity fixes in this PR also close a few compatibility gaps that matter for real cache scans:
- unresolved env vars now remain literal instead of being dropped from configured paths
- explicit cache paths passed to `scan_dir` now behave like Python's `expanduser().resolve()` flow
- malformed singular cache directories such as `model--repo` are now treated as invalid instead of being accepted during scanning

## Validation
- `cargo fmt`
- `cargo test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ It aims to be compatible with the huggingface_hub python package, but only imple
 
 [dependencies]
 futures = { version = "0.3.28", optional = true }
+comfy-table = { version = "4.1.1", optional = true }
 dirs = "6"
 http = { version = "1.0.0", optional = true }
 indicatif = { version = "0.18", optional = true }
@@ -34,6 +35,7 @@ thiserror = { version = "2", optional = true }
 tokio = { version = "1.29.1", optional = true, features = ["fs", "macros"] }
 ureq = { version = "3", optional = true, features = ["json", "socks-proxy"] }
 native-tls = { version = "0.2.12", optional = true }
+walkdir = { version = "2.5.0", optional = true }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
 version = "0.61"
@@ -46,6 +48,8 @@ optional = true
 
 [features]
 default = ["default-tls", "tokio", "ureq"]
+cache-manager = ["dep:thiserror", "dep:walkdir"]
+cache-manager-display = ["dep:comfy-table", "cache-manager"]
 # These features are only relevant when used with the `tokio` feature, but this might change in the future.
 default-tls = ["native-tls"]
 native-tls = [
@@ -89,3 +93,8 @@ ureq = [
 hex-literal = "1.0"
 sha2 = "0.10"
 tokio-test = "0.4.2"
+
+[[example]]
+name = "cache-manager"
+path = "examples/cache-manager.rs"
+required-features = ["cache-manager"]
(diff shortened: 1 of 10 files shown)


i386:
@Narsil would you mind taking a quick look at this PR please?

Narsil:
I'm not at hf anymore ask @McPatate

assafvayner:
Hi @i386 

I have been working on an alternative huggingface hub rust library that is currently in [huggingface/huggingface_hub_rust](https://github.com/huggingface/huggingface_hub_rust).

This library includes a lot of updates, including the repo search, cache implementation and progress tracking. @McPatate and I are currently thinking to bring that implementation into hf-hub as a major version release soon, if there is interest in this now, we will try to expedite the process.

We are intent on including a cache that can inter-operate with the huggingface_hub python library cache implementation (which I understand your implementation is based on the python version, but we would need some tests).

So with all that, let us know if you have thoughts and if you prefer an explicit cache interface like you created?

Check the README on [huggingface/huggingface_hub_rust](https://github.com/huggingface/huggingface_hub_rust) let me know if that would help you.

i386:
@assafvayner thanks for the reply! Yes, we'd absolutely need interoperability with the official cache. Are you also thinking about XET too? If you're open to it I'd love to contribute

i386:
@assafvayner this is working quite well for us https://github.com/Mesh-LLM/mesh-llm/pull/292 - thank you!

McPatate:
@i386 note that we're in the process of merging the two crates! We'll keep hf-hub's repo, but port over huggingface_hub_rs's code here, so you won't need to do a second migration 😄

Should only be a matter of updating the Cargo.toml back to hf-hub :)

PR #149, opened 2026-04-14
feat: add download and API benchmarks for sync

McPatate:
Preparing for `1.0`, thought it might be worthwhile adding a small benchmark to check for regressions. Lmk what you think, can discard it if we feel it doesn't bring value.

### DIFF
--- a/benches/sync_api.rs
+++ b/benches/sync_api.rs
@@ -1,0 +1,190 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use hf_hub::api::sync::ApiBuilder;
+use hf_hub::{Cache, Repo, RepoType};
+use std::path::PathBuf;
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+fn bench_cache_dir() -> PathBuf {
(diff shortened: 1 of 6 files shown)


benches/sync_api.rs
```diff
@@ -0,0 +1,190 @@
...
+// ---------------------------------------------------------------------------
+
+fn bench_cache_dir() -> PathBuf {
+    let mut path = std::env::temp_dir();
+    path.push(format!("hf_hub_bench_{}", std::process::id()));
+    std::fs::create_dir_all(&path).ok();
+    path
+}
+
+fn fresh_cache_dir() -> PathBuf {
```
assafvayner: I recommend to use https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html instead (`tempfile::tempdir()`)
so the dirs get cleaned implicitly

assafvayner (approved):
Approving to unblock, looks good if you can get it working. Looks like there's issues in github actions trying to run.

benches/sync_api.rs
McPatate: thought of that but was thinking it was a bit overkill, I'll add it, it's just a dev-dep anyways

PR #161, opened 2026-05-13
feat: streaming xet upload sessions for repos and buckets

znation:
## Motivation

The existing repo and bucket upload APIs require every file in a commit to be
buffered in memory (`AddSource::Bytes`) or staged on local disk
(`AddSource::File` / `BucketUpload { local, .. }`) until commit time. That
makes the total commit payload bounded by local RAM or disk.

For workloads that generate output incrementally and may exceed either bound —
rendered tile pyramids, sharded training outputs, streaming exports — there's
currently no public seam to upload-then-reference, even though the internal
`xet_upload_inner` already does exactly that with a `pub(crate)` API.

This PR opens that seam.

### Repo side
- **`AddSource::Lfs { sha256_oid, size }`** — a new `AddSource` variant that
  references an already-uploaded LFS blob by SHA-256 OID. `create_commit`
  short-circuits preupload, SHA-256 recomputation, and xet transfer for these,
  emitting an `lfsFile` NDJSON entry directly.
- **`HFRepository::open_xet_upload_session(revision)`** / blocking counterpart
  — opens an `XetUploadSession` whose `upload_bytes` / `upload_file` push to
  xet CAS as the caller produces content. `finish()` seals the CAS state and
  returns `HashMap<String, UploadedXetFile>` carrying `(sha256_oid, size,
  xet_hash)` per path, ready to feed into `AddSource::Lfs`.
- **`CommitOperation::add_lfs(path, oid, size)`** convenience helper.

### Bucket side
- **`HFBucket::open_xet_upload_session()`** / blocking counterpart — same
  shape. `finish()` returns descriptors keyed by remote path;
  `finish_and_register()` chains a `batch_operations` call to make the files
  visible in one step.

## Caller pattern

```rust
// Repos
let session = repo.open_xet_upload_session(None).await?;
for (path, bytes) in produce_tiles() {
    session.upload_bytes(path, bytes).await?; // bytes can be dropped after this
}
let uploaded = session.finish().await?;
let ops: Vec<CommitOperation> = uploaded
    .into_iter()
    .map(|(p, f)| CommitOperation::add_lfs(p, f.sha256_oid, f.size))
    .collect();
repo.create_commit().operations(ops).commit_message("…").send().await?;

// Buckets
let session = bucket.open_xet_upload_session().await?;
for (path, bytes) in produce_tiles() {
    session.upload_bytes(path, bytes).await?;
}
session.finish_and_register().await?;
```

The caller holds only `(path, sha256_oid, size, xet_hash)` triples between
upload and finalize — bounded by O(file count), not O(bytes).

## Internals

- Extracted the xet session bring-up out of `xet_upload_inner` into a reusable
  `pub(crate) new_xet_upload_commit(...)` helper that both the existing
  batched flow and the new streaming sessions share, so the token fetch +
  session-poisoning fallback logic stays in one place.
- `xet_upload_inner` and the existing matches on `AddSource` reject
  `AddSource::Lfs` defensively — it's routed via `lfs_uploaded` upstream of
  preupload and never reaches the inline-base64 or xet-transfer arms. Errors
  there indicate a logic bug, not a user error.
- The streaming sessions use the same `XetUploadCommit` / `Sha256Policy::Compute`
  primitives that the batched flow already uses, so per-file dedup, retry, and
  session-poisoning behavior are unchanged.

## Test plan

- Unit tests added for the new `AddSource::Lfs` and `CommitOperation::add_lfs`
  constructors.
- Existing unit + doc tests pass with and without the `blocking` feature
  (`cargo test -p hf-hub` / `cargo test -p hf-hub --features blocking`,
  163 tests).
- `cargo clippy -p hf-hub --features blocking --all-targets` is clean on the
  changed files.
- End-to-end streaming-upload behavior needs a live Hub and will be exercised
  by dependent crates; happy to add integration tests here too if you'd
  rather they live in this repo.

## Notes

- Driven by a real downstream use case: rendering Hilbert-curve binary
  visualization tile pyramids whose total size routinely exceeds local disk.
  See [arbvis](https://github.com/znation/arbvis) — the streaming sink there
  currently has to fall back to a TempDir + `add_file` flow with a comment
  pointing here.
- The PR is intentionally minimal — no progress reporting on the streaming
  sessions yet, and `finish` waits for all uploads to drain rather than
  exposing per-file completion. Both are easy follow-ups if there's interest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hf-hub/src/buckets/mod.rs
+++ b/hf-hub/src/buckets/mod.rs
@@ -14,6 +14,11 @@
 //! - Use [`sync`] for one-way directory mirroring between a local folder and a bucket prefix.
 
 pub mod sync;
+pub mod xet_session;
+
+pub use xet_session::BucketXetUploadSession;
+#[cfg(feature = "blocking")]
+pub use xet_session::BucketXetUploadSessionSync;
 
 use std::path::PathBuf;
 
--- a/hf-hub/src/buckets/mod.rs
+++ b/hf-hub/src/buckets/mod.rs
@@ -16,10 +16,6 @@
 pub mod sync;
 pub mod xet_session;
 
-pub use xet_session::BucketXetUploadSession;
-#[cfg(feature = "blocking")]
-pub use xet_session::BucketXetUploadSessionSync;
-
 use std::path::PathBuf;
 
 use bon::bon;
@@ -28,6 +24,9 @@ use futures::Stream;
 use futures::StreamExt;
 use serde::{Deserialize, Serialize};
 use url::Url;
+pub use xet_session::BucketXetUploadSession;
+#[cfg(feature = "blocking")]
+pub use xet_session::BucketXetUploadSessionSync;
 
 use crate::client::HFClient;
 use crate::error::{HFError, HFResult, NotFoundContext};
--- a/hf-hub/src/buckets/xet_session.rs
+++ b/hf-hub/src/buckets/xet_session.rs
@@ -24,7 +24,7 @@ use std::sync::Mutex;
 
 use xet::xet_session::{Sha256Policy, UniqueID, XetFileUpload};
 
-use super::{HFBucket, BucketAddFile};
+use super::{BucketAddFile, HFBucket};
 use crate::error::{HFError, HFResult, NotFoundContext, XetOperation};
 use crate::repository::UploadedXetFile;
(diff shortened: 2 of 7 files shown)


znation:
Sorry for the spam, I think my AI slop doesn't make sense.

PR #165, opened 2026-05-19
ci(security): route CI through HF internal registry proxies

assafvayner:
## Summary

Routes cargo and pip fetches in CI through the HuggingFace internal package registry proxies described in [this Notion doc](https://www.notion.so/huggingface2/Internal-Package-Registry-Proxies-3641384ebcac8199b610c4f98f376986). The proxies hide package versions younger than 3 days, which catches the typosquatting / account-takeover window of most public-registry supply-chain attacks before any version reaches our build hosts.

## What changes

- Every Linux job that installs packages runs `curl -sSL https://registries.huggingface.tech/setup.sh | bash` right after the Rust toolchain install. The bootstrap script auto-detects cargo / pip / uv / npm and writes the right config idempotently.
- `rust.yml`: every Linux job (fmt, lint, docs, audit, plus the Linux row of build-and-test) moves to `runs-on: { group: aws-general-8-plus }`. Windows and macOS rows in the build-and-test matrix stay on GitHub-hosted runners and skip the setup step (the proxies' ALB is `scheme: internal`).
- `rust-release.yml`: publish job moves to `aws-general-8-plus` and runs the setup script before `cargo publish`. `cargo publish`'s upload still hits crates.io directly; the mirror only affects fetches for the verification build.
- `benchmarks.yml`: benchmark job moves to `aws-general-8-plus` and runs the setup script.
- `benchmark-trigger.yml` is unchanged (it only calls the GitHub API, no package installs).

## Constraints worth flagging

- **GitHub-hosted runners cannot reach the proxies** (the ALB is `scheme: internal`). Verified empirically — every curl from `ubuntu-latest`/`windows-latest`/`macOS-latest` times out on TCP connect. The Windows and macOS rows in the test matrix stay on GitHub-hosted runners and skip the setup step; supply-chain protection only applies to the Linux jobs in this PR. Open question: do we want to drop cross-platform CI entirely to close that gap, or accept the asymmetry?
- **Cache key for `rust-release.yml`** was `ubuntu-latest-cargo-registry-…`; bumped it to `aws-general-8-plus-cargo-registry-…` so we don't pull stale cache entries onto the new runner pool.
- **Rustup downloads are not proxied** — `dtolnay/rust-toolchain` still fetches from `static.rust-lang.org`. The Notion doc doesn't list a rustup mirror; we rely on SHA-pinned action versions there.

## Test plan

- [x] All 7 checks pass on the in-VPC `aws-general-8-plus` runners (run `26127462882`).
- [x] Linux jobs are noticeably faster on the new runners (Format 5m13s → 13s, Build & Test 7m42s → 3m34s).
- [ ] Tag a pre-release to exercise the `rust-release.yml` path before relying on it for an actual release.


McPatate (approved):
👌🏻

PR #170, opened 2026-05-27
Support resumable downloads

i386:
## Summary

Adds resumable download support for repository downloads across the normal HTTP/LFS path and the Xet path.

- Resume partial cache blob downloads from existing `.incomplete` files with HTTP `Range` requests.
- Resume `local_dir` HTTP/LFS downloads from partial destination files.
- Validate `206 Partial Content` and `Content-Range`; fall back to a clean restart when the server ignores or rejects the requested range.
- Resume Xet cache and local downloads by appending ranged Xet stream bytes to the existing partial file.
- Reuse shared partial-download state handling for complete, resumable, oversized, missing, and forced-download cases.

## Why

The previous implementation preserved partial files, but restarted downloads by truncating them on retry. Large model downloads, especially GGUF/LFS/Xet artifacts, therefore had to start from byte 0 after interruption instead of continuing from the already-written bytes.

## Validation

- `cargo test -p hf-hub --lib` (`172 passed`)
- `cargo fmt --check` with nightly rustfmt
- `git diff --check`

The added tests cover HTTP/LFS cache resume, HTTP/LFS `local_dir` resume, invalid range fallback, shared partial-state classification, and Xet append/resume behavior.


i386:
@assafvayner could I please get an eye on this one as well

assafvayner:
I need to double check, but this is not currently supported/supportable AFAIK. Xet files are not guaranteed to be written in order and can contain gaps until completed (even if they are now, no long term guarantee that this behavior won't change).

I recommend instead to experiment with enabling the xet cache https://huggingface.co/docs/huggingface_hub/main/en/guides/manage-cache#chunk-based-caching-xet

hf-hub using hf-xet should accept the same env var configurations right now. If you can test if it solves your issues and let me know if you have a different interface preference we can implement it (both in hf-hub and/or hf-xet).

CC @McPatate if you have any opinion

PR #172, opened 2026-05-27
Match Python cache fallback when symlinks are unavailable

i386:
## Summary

Mirror Python `huggingface_hub` cache behavior for platforms where snapshot symlinks cannot be created, especially Windows without Developer Mode/admin symlink privileges.

- Try to create the snapshot pointer as a symlink on Windows instead of unconditionally copying.
- If symlink creation fails, move a freshly downloaded blob into the snapshot path so a new download does not leave both `blobs/<etag>` and `snapshots/<commit>/<filename>` as full-size files.
- Continue copying existing blobs into snapshots when symlinks are unavailable, because an existing blob may already be referenced by other snapshots.
- Reuse an existing regular snapshot file after metadata lookup when the cache is already in degraded no-symlink mode and the blob path is absent.

## Python reference

This follows the Python `huggingface_hub` cache fallback behavior:

- Symlink support is probed per cache directory and Windows users without symlink support are warned that cache behavior is degraded: https://github.com/huggingface/huggingface_hub/blob/07a6ccc819fd531c3fdb984954133c27271a412a/src/huggingface_hub/file_download.py#L2354-L2452
- `_create_symlink` documents the degraded behavior: newly downloaded blobs are moved into the destination, while existing blobs are copied to avoid breaking other snapshots: https://github.com/huggingface/huggingface_hub/blob/07a6ccc819fd531c3fdb984954133c27271a412a/src/huggingface_hub/file_download.py#L608-L634
- The actual fallback branch moves when `new_blob=True` and copies otherwise: https://github.com/huggingface/huggingface_hub/blob/07a6ccc819fd531c3fdb984954133c27271a412a/src/huggingface_hub/file_download.py#L693-L699

## Why

The Python implementation avoids keeping both the extensionless blob and the named snapshot file for a fresh download when symlinks are unavailable. The current Rust implementation always copies on Windows, which can double disk usage for large model files.

## Tests

Added focused regression coverage for the degraded no-symlink path:

- `pointer_fallback_moves_new_blob` verifies fresh blobs are moved to the snapshot path and the blob path is removed.
- `pointer_fallback_copies_existing_blob` verifies existing blobs are copied so other snapshot references remain valid.
- `cache_download_reuses_regular_snapshot_when_blob_is_absent` verifies a cache already in degraded no-symlink mode reuses the regular snapshot file instead of downloading again.

## Validation

- `cargo fmt --check` with nightly rustfmt
- `cargo test -p hf-hub --lib` (`167 passed`)
- `git diff --check`


i386:
@assafvayner @McPatate could I please get a review on this PR and also https://github.com/huggingface/hf-hub/pull/170 ?

We have to maintain a fork for [Mesh LLM](https://github.com/Mesh-LLM/mesh-llm/) and it is blocking our release for cargo. Thanks!

assafvayner:
Implementation plan for the follow-up fix:

1. Preserve the existing snapshot reuse behavior, but update the requested revision ref before returning an already-present snapshot.
2. Add a regression test for downloading by commit hash, then by branch, followed by an offline branch lookup.
3. Run cargo +nightly fmt, cargo clippy -p hf-hub --all-features -- -D warnings, and cargo test -p hf-hub.

assafvayner:
Hey @i386 I'm replacing this PR with #190 (based on this PR, but I preferred not to push to your fork).

waiting for an ack from @McPatate then should be able to merge and release soon after.

Thank you for raising this and sorry for the delay

assafvayner:
replaced by #190.

i386:
@assafvayner thank you for this! and feel free to push to my branches. we have a branch where we combine all of our changes.

PR #174, opened 2026-06-09
fix(error): surface Hub server message in HFError Display

assafvayner:
## What

HTTP-derived `HFError` variants captured the Hub's error explanation in `HttpErrorContext.server_message` but dropped it from their `Display`, so a bubbled-up rejection (e.g. a >1000-file commit → 400/413 → `HFError::Http`) showed only status + url + ids, not the reason.

`hf-hub/src/error.rs` only: replace `format_http_suffix` with `format_http_detail`, which leads with `server_message` then appends `(url=…, request_id=…, error_code=…)`, omitting absent pieces. Applied to `Http`, `AuthRequired`, `Forbidden`, `Conflict`, `RateLimited`.

Live 403 example:
```
Forbidden: You don't have the rights to create a model under the namespace "openai-community" (url=…, request_id=Root=1-…)
```

No client-side count check / splitting (matches `huggingface_hub` bubble-up); only the `Display` string changes — not a breaking API change.

## Testing
fmt + clippy clean; `cargo test -p hf-hub --all-features` → 188 passed (incl. new Display tests); manually verified the live 403 above.

### DIFF
--- a/hf-hub/src/error.rs
+++ b/hf-hub/src/error.rs
@@ -58,13 +58,29 @@ pub(crate) fn parse_retry_after(headers: &HeaderMap) -> Option<Duration> {
     raw.trim().parse::<u64>().ok().map(Duration::from_secs)
 }
 
-fn format_http_suffix(ctx: &HttpErrorContext) -> String {
-    match (&ctx.request_id, &ctx.error_code) {
-        (Some(rid), Some(code)) => format!(" (request_id={rid}, error_code={code})"),
-        (Some(rid), None) => format!(" (request_id={rid})"),
-        (None, Some(code)) => format!(" (error_code={code})"),
-        (None, None) => String::new(),
+/// Render the human-readable tail shared by every HTTP-derived [`HFError`] variant.
+///
+/// Leads with the server's explanation (`server_message`) when present, then appends a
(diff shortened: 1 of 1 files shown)


hf-hub/src/error.rs
```diff
@@ -58,13 +58,29 @@ pub(crate) fn parse_retry_after(headers: &HeaderMap) -> Option<Duration> {
...
-    match (&ctx.request_id, &ctx.error_code) {
-        (Some(rid), Some(code)) => format!(" (request_id={rid}, error_code={code})"),
-        (Some(rid), None) => format!(" (request_id={rid})"),
-        (None, Some(code)) => format!(" (error_code={code})"),
-        (None, None) => String::new(),
+/// Render the human-readable tail shared by every HTTP-derived [`HFError`] variant.
+///
+/// Leads with the server's explanation (`server_message`) when present, then appends a
+/// parenthetical of `url`, `request_id`, and `error_code`. Each piece is omitted when absent,
+/// so a context carrying only a URL renders as ` (url=…)` with no dangling separators.
```
McPatate: This is quite redundant with code that is self-explanatory imo. Perhaps would be better to remove?

McPatate (approved):
Small nit on the comment, feel free to ignore

PR #178, opened 2026-07-03
fix: encode revisions as single path segments, follow relative redirects; add split_id helper

assafvayner:
Quality-of-life batch surfaced while migrating downstream consumers onto `hf-hub`. Two bug fixes, three ergonomic additions (`split_id`, anonymous clients, `HFError::status_code`), and doc notes.

### 1. Encode revisions as single URL path segments
Slash-containing revisions (`refs/pr/N`, `refs/convert/parquet`) were interpolated raw into resolve/api URLs, so the Hub read them as extra path segments and returned **404** for every request against them.

New `encode_path_segment()` in `client.rs` percent-encodes a value as exactly one path segment (`refs/pr/21` -> `refs%2Fpr%2F21`), mirroring `huggingface_hub`'s `quote(revision, safe="")`. Applied at all 16 revision/branch/tag/compare interpolation sites: `download_url`, `revision`, `tree`, `paths-info`, `commits`, `compare`, `branch`, `tag`, `commit`, `preupload`, and the xet-token URLs.

**Repro / evidence:** surfaced by candle's `hub::tests::test_dataset` (resolves `refs/convert/parquet`) during consumer validation; verified live via `hfrs download --revision refs/pr/21`.

### 2. Follow relative redirects on metadata HEAD requests
File-metadata resolution HEADs via the no-redirect client. When a repo id differs only in casing (`snowflake/snowflake-arctic-embed-xs` vs canonical `Snowflake/...`), the Hub 307s to the canonical repo path with **no ETag**, so the old code surfaced a misleading "missing ETag header" error.

New `head_with_relative_redirects()` + `is_relative_location()` in `client.rs`, used at both HEAD sites in `repository/download.rs`. Only relative redirects (Location without a host) are followed; absolute/CDN redirects are returned as-is since their headers already carry the linked-file metadata. Mirrors `huggingface_hub`'s `follow_relative_redirects`.

**Repro / evidence:** surfaced by fastembed's test suite (non-canonical casing); verified live via `hfrs download snowflake/snowflake-arctic-embed-xs`.

## New helper: `split_id()`
`hf_hub::split_id(id) -> (owner, name)` splits a repo/bucket id on its first `/`, yielding an empty owner for short-form ids (`"gpt2"` -> `("", "gpt2")`) — feeds straight into `client.model(owner, name)` and the other handle constructors.

Rationale: every migrated consumer (tokenizers, candle, TEI, vLLM, sglang, fastembed, modelexpress, dynamo) hand-wrote the same `split_once('/')` match, and hfrs carried ~13 copies. Refactored those copies in `commands/{download,upload}`, `repos/{branch,delete_files,file_metadata,tag,settings}`, and `{models,datasets,spaces}/info` to use the helper. `parse_bucket_id` deliberately keeps its own logic (it validates and errors on a missing slash rather than defaulting to an empty owner).

## Anonymous client support
`HFClient::builder().anonymous().build()` builds a client that sends requests unauthenticated even when ambient credentials exist (`HF_TOKEN` env var, `HF_TOKEN_PATH`, cached `$HF_HOME/token`) — the per-client equivalent of the process-wide `HF_HUB_DISABLE_IMPLICIT_TOKEN` env var, which was previously the only escape hatch (mistral.rs friction).

Semantics: `.anonymous()` wins over all implicit token sources (that's its whole point); setting both `.token(...)` and `.anonymous()` is contradictory and fails `build()`/`build_sync()` with `HFError::InvalidParameter` rather than silently picking one. Documented on the builder and in the crate-level client docs.

## `HFError::status_code()`
`err.status_code() -> Option<reqwest::StatusCode>` extracts the HTTP status the server actually returned, replacing manual digging through `HttpErrorContext` (mistral.rs friction). `StatusCode` was already public surface via `HttpErrorContext::status`.

Covers the context-bearing variants (`Http`, `AuthRequired`, `Forbidden`, `Conflict`, `RateLimited`), the not-found family when their optional response context was captured, and `Request` errors whose `reqwest` source knows a status. No status is ever synthesized — a `RepoNotFound` built without a response context returns `None` rather than a canonical 404, and purely local errors (I/O, JSON, cache) always return `None`.

## Doc notes
- `ModelInfo`/`RepoInfo` optional `siblings`/`sha`: show the `.iter().flatten()` / `.unwrap_or_default()` idiom.
- `download_file().local_files_only(true)` as the replacement for 0.x `Cache::get`, and matching `HFError::LocalEntryNotFound` to distinguish "not cached" from a real `EntryNotFound`/transport error.

## Verification
- `cargo test -p hf-hub`: 213 unit + 27 doc tests pass (216 with `--all-features`); includes new tests for `split_id`, `encode_path_segment`/`is_relative_location`, builder token resolution with `anonymous()` + `HF_TOKEN`/token-file in env, and `status_code()` across variant families
- `cargo test -p hfrs`: 73 pass; 24 remaining failures are all `HF_TEST_WRITE=1`-gated write/xet tests (environmental, unchanged by this PR)
- `cargo clippy --all-features -- -D warnings`: clean on hf-hub and hfrs
- `cargo +nightly fmt`: clean
- `./scripts/verify_wasm.sh`: green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,6 +1131,7 @@ dependencies = [
  "hf-xet",
  "hyper",
  "pathdiff",
+ "percent-encoding",
  "reqwest",
  "serde",
  "serde_json",
--- a/hf-hub/Cargo.toml
+++ b/hf-hub/Cargo.toml
@@ -22,6 +22,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 thiserror = "2"
 url = "2"
+percent-encoding = "2"
 futures = "0.3"
 bon = "3"
 globset = "0.4"
--- a/hf-hub/src/client.rs
+++ b/hf-hub/src/client.rs
@@ -29,6 +29,19 @@ pub(crate) fn append_path_segments(url: &mut Url, path: &str) -> HFResult<()> {
     Ok(())
 }
 
+/// Percent-encode `value` as a single URL path segment.
+///
+/// Unlike [`append_path_segments`], slashes are escaped rather than treated as
+/// separators, so refs like `refs/pr/21` or `refs/convert/parquet` occupy
+/// exactly one segment (`refs%2Fpr%2F21`) — the Hub 404s on the unescaped
+/// form. Mirrors `huggingface_hub`'s `quote(value, safe="")`. Use for
+/// revisions, branch names, and tag names interpolated into URL paths.
+pub(crate) fn encode_path_segment(value: &str) -> String {
+    let mut url = Url::parse("https://hf.co").expect("static base URL parses");
+    url.path_segments_mut().expect("https URL can be a base").push(value);
+    url.path()[1..].to_string()
+}
+
 /// Async client for the Hugging Face Hub API.
 ///
 /// `HFClient` wraps an `Arc<HFClientInner>` so it is cheap to clone — all clones
@@ -345,7 +358,8 @@ impl HFClient {
     /// `url_prefix` is `""` for models or `"<plural>/"` for the other repo types. Pass
     /// [`crate::repository::RepoType::url_prefix`] for the concrete repo type at the call
     /// site. `filename` is percent-encoded path-segment by path-segment, so values like
-    /// `subdir/file name #1.bin` produce a correctly escaped URL.
+    /// `subdir/file name #1.bin` produce a correctly escaped URL. `revision` is encoded
+    /// as a single segment, so `refs/pr/21` becomes `refs%2Fpr%2F21`.
     pub(crate) fn download_url(
         &self,
         url_prefix: &str,
@@ -353,7 +367,7 @@ impl HFClient {
         revision: &str,
         filename: &str,
     ) -> HFResult<String> {
-        let base = format!("{}/{}{}/resolve/{}", self.endpoint(), url_prefix, repo_id, revision);
+        let base = format!("{}/{}{}/resolve/{}", self.endpoint(), url_prefix, repo_id, encode_path_segment(revision));
         let mut url = Url::parse(&base)?;
         append_path_segments(&mut url, filename)?;
         Ok(url.into())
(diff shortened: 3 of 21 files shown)


assafvayner:
## Implementation plan

### Scope
Two already-verified bug fixes plus one ergonomic helper and doc notes, split into four focused commits.

### Commits
1. `fix: encode revisions as single URL path segments` — `encode_path_segment()` in `client.rs` + 16 call sites across `download.rs`/`commits.rs`/`listing.rs`/`mod.rs`/`upload.rs`/`xet.rs`.
2. `fix: follow relative redirects on metadata HEAD requests` — `head_with_relative_redirects()` + `is_relative_location()` in `client.rs`, wired into both HEAD sites in `repository/download.rs`.
3. `feat: add split_id helper and adopt it across hfrs commands` — new `hf_hub::split_id`, re-exported at the crate root, with unit tests and doctests; refactors ~13 hand-written `split_once('/')` matches in hfrs.
4. `docs: note optional siblings/sha access and cache-only lookups` — rustdoc on `ModelInfo` and `download_file`, plus a crate-doc `split_id` example.

### Testing strategy
- Unit tests for `encode_path_segment`, `is_relative_location`, and `split_id` (all in `client.rs` `#[cfg(test)]`).
- Doctests on `split_id`, `ModelInfo`, and `download_file` exercise the documented idioms.
- hfrs read-only comparison tests cover the refactored command paths.
- Manual live verification of both fixes via `hfrs download` (see PR description).
- Full gate: `cargo test`, `cargo clippy --all-features -D warnings`, `cargo +nightly fmt`, `./scripts/verify_wasm.sh`.

### Deliberately left alone
- hfrs `parse_bucket_id`: validates and errors on a missing slash, so its semantics don't match `split_id`'s empty-owner default.
- hf-hub internal `split_repo_id`: returns `(Option<&str>, &str)`, a different shape from `split_id`.

PR #180, opened 2026-07-03
fix: box large async-method futures to keep consumer future depth shallow

assafvayner:
## Motivation

During pre-release consumer validation, migrating vLLM's `rust/` workspace to hf-hub 1.0 broke the build of **three downstream targets that don't even import hf-hub** with `error: queries overflow the depth limit!`. Simply awaiting `download_file().send()` inside vLLM's async call chains embedded hf-hub's download state machine into every enclosing async fn, growing and deepening the future type until rustc's recursion/query depth gave out. vLLM had to `Box::pin` at its integration boundary to compile at all — and every non-trivial async consumer would hit the same wall and blame their own code. This fixes it at the source.

## What changed

The heavyweight builders now box their inner future behind a type-erased pointer via a new `pub(crate)` helper, `crate::util::boxed_future` — `BoxFuture` (`+ Send`) on native, `LocalBoxFuture` on `wasm32-unknown-unknown` where the browser reqwest backend produces `!Send` futures. The `#[bon]` method signatures are unchanged; only the method bodies now delegate through the box, so the future returned by `send()` is a small, shallow state machine holding the builder parameters plus one pinned pointer.

## Future sizes before/after (`size_of_val` of the built-but-not-awaited `send()` future)

| Builder | Before | After |
|---|---:|---:|
| `HFRepository::download_file` | 27,760 B | 336 B |
| `HFRepository::download_file_stream` | 27,048 B | 320 B |
| `HFRepository::download_file_to_bytes` | 27,152 B | 320 B |
| `HFRepository::snapshot_download` | 27,688 B | 456 B |
| `HFRepository::create_commit` | 28,144 B | 480 B |
| `HFRepository::upload_file` | 28,648 B | 600 B |
| `HFRepository::upload_folder` | 28,848 B | 704 B |
| `HFRepository::delete_file` | 28,128 B | 288 B |
| `HFRepository::delete_folder` | 30,248 B | 288 B |
| `HFBucket::download_file_stream` | 26,680 B | 176 B |
| `HFBucket::upload_source_files` | 26,896 B | 176 B |
| `HFBucket::upload_files` | 26,864 B | 176 B |
| `HFBucket::download_files` | 26,592 B | 176 B |
| `HFBucket::sync` | 27,880 B | 408 B |

**Deliberately not boxed** (boxing would be pure overhead on already-small futures):

| Builder | Size |
|---|---:|
| `HFBucket::batch_operations` | 2,288 B |
| `HFRepository::info` (control) | 2,152 B |
| `HFBucket::info` (control) | 1,920 B |

`info`/`exists`/refs/listing-style builders are all in the same ~2 KB class and were left alone.

## Notes

- **Regression test**: `hf-hub/tests/future_size.rs` asserts every boxed `send()` future stays under 2 KB (headroom for new parameters, still catches an accidentally unboxed 27 KB state machine), plus a consumer-shaped test with three nested async fns awaiting `download_file().send()` — the exact shape that broke vLLM — asserting the nested future stays shallow and `Send`.
- **HRTB fallout fixed**: Send-boxing the snapshot download future exposed a latent rustc "implementation of `FnOnce` is not general enough" limitation in two closure-returning-future sites (`head_futs` in `snapshot_download_impl`, and `download_concurrently`). Both are restructured into plain loops pushing futures into a `Vec` — behaviorally identical. This also means the snapshot future now demonstrably passes a `Send` check, which it previously never had to.
- **Blocking mirrors**: unaffected — the `*Sync` wrappers `block_on` the same builders; confirmed compiling and passing under `--all-features`.
- **wasm**: `boxed_future` drops the `Send` bound on wasm (`LocalBoxFuture`); `./scripts/verify_wasm.sh` is green.

## Verification

- `cargo test -p hf-hub` — green (includes the new size regression tests)
- `cargo clippy -p hf-hub --all-features -- -D warnings` — clean
- `cargo +nightly fmt` — applied
- `./scripts/verify_wasm.sh` — green
- `cargo check --workspace --all-features` — green (hfrs, integration-tests, examples, benches)


assafvayner:
## Implementation plan

1. **Measure first** — add a probe (`hf-hub/tests/future_size.rs`) that builds each heavyweight builder's `send()` future offline (no network; `HFClient::builder().endpoint("http://localhost:9")…`) and prints `size_of_val`. Candidates: `download_file`, `download_file_stream`, `download_file_to_bytes`, `snapshot_download`, `create_commit`, `upload_file`, `upload_folder`, `delete_file`, `delete_folder`, `HFBucket::{download_file_stream, upload_source_files, upload_files, download_files, batch_operations, sync}`, with `info` builders as small controls. Result: everything except `batch_operations` (~2.3 KB) and the controls (~2 KB) sat at 26-30 KB.
2. **Boxing helper** — `crate::util::boxed_future` in a new `hf-hub/src/util.rs`: `Box::pin` into `futures::future::BoxFuture` (`F: Send`) on native, `LocalBoxFuture` on wasm where reqwest futures are `!Send`. cfg selection keeps the bon method signatures and the wasm cfg structure untouched.
3. **Apply** — builders that delegate to a private `*_impl` wrap that call (`boxed_future(self.x_impl(params)).await`); builders with inline bodies (`HFBucket::{download_file_stream, upload_files, download_files}`) wrap the body in `boxed_future(async move { … }).await`. `batch_operations` and the info/exists/listing class left unboxed — already small, boxing is pure overhead.
4. **HRTB fix** — the `Send` bound on the boxed snapshot future surfaced rustc's "implementation of `FnOnce` is not general enough" limitation at two closure-returning-future sites (`head_futs`, `download_concurrently`); restructured both into loops pushing futures into a `Vec`, behaviorally identical.
5. **Regression tests** — convert the probe into assertions (each boxed `send()` future < 2048 B) plus a consumer-shaped test: three nested async fns awaiting `download_file().send()` (the vLLM pattern), asserting the nested future stays shallow and is `Send`.
6. **Blocking mirrors** — no change needed; `*Sync` wrappers `block_on` the same builders. Verified via `--all-features` builds/tests.

## Testing strategy

- Unit/size: `cargo test -p hf-hub` (includes the new `future_size` tests, before/after table in the PR body).
- Lint/format: `cargo clippy -p hf-hub --all-features -- -D warnings`, `cargo +nightly fmt`.
- wasm: `./scripts/verify_wasm.sh`.
- Consumers: `cargo check --workspace --all-features` (hfrs, integration-tests, examples, benches).
- Live-network integration tests unchanged and not required: the change is purely how futures are allocated, not what they do.

PR #181, opened 2026-07-03
feat: make hf-xet optional behind a `xet` feature

assafvayner:
## What

Makes the `hf-xet` dependency **optional**, gated behind a new `xet` cargo feature (**on by default**). When the feature is disabled, `hf-xet` is dropped from the dependency tree and any transfer that would use xet fails with a descriptive `HFError::XetFeatureDisabled` error — there is **no silent fallback** to plain HTTP.

## Motivation

Consumer validation surfaced real breakage from hf-hub 1.0's unconditional `hf-xet` dependency:

- `hf-xet` (via xet-core) floors **tokio at `^1.49`** (it uses `Handle::id()`, stabilized in tokio 1.49).
- **sglang's router** and **NVIDIA dynamo** both pin `tokio = "=1.48.0"` (dynamo ships this pin in its published crates). Against hf-hub 1.0 this makes cargo resolution *impossible* — no tokio version satisfies both.
- sglang only pulls two small, git-stored JSON files (`tokenizer.json`, `tokenizer_config.json`). These are **not** xet-backed, so xet buys them nothing while forcing the tokio floor and a large transitive tree onto them.

Making `hf-xet` optional lets metadata / small-file / non-xet consumers resolve freely and shrink their dependency tree, while xet stays available (and default-on) for consumers that want fast large-file transfers. This mirrors Python `huggingface_hub`, where `hf_xet` is an optional extra.

## Design decision: fail loudly, no HTTP fallback

Per the agreed design, when the `xet` feature is **off**, xet-requiring transfers **fail with a clear runtime error** rather than degrading. New error variant:

```rust
HFError::XetFeatureDisabled { operation: XetOperation }
```

Where the errors sit (the seams where the xet path would have been taken):

| Path | Seam |
|------|------|
| Download (cache) | `download_file_to_cache_network` — when the metadata HEAD carries `x-xet-hash` |
| Download (local_dir) | `download_file_to_local_dir` — when the HEAD carries `x-xet-hash` |
| Download (stream) | `download_file_stream_impl` — when `paths-info`/HEAD reports a xet hash |
| Snapshot download | the `xet_batch_fut` in both `local_dir` and cache branches — errors if any file in the set is xet-backed |
| Upload | `upload_lfs_files_via_xet` — after LFS batch negotiation |
| Buckets | `download_file_stream`, `download_files`, `upload_files`, `upload_source_files`, and `sync` download plans (buckets are xet-backed, so these error wholesale) |

### Upload: steering the LFS negotiation

Following the "advertise only non-xet transfers" approach: when `xet` is off, the LFS batch request advertises only `["basic", "multipart"]` instead of `["basic", "multipart", "xet"]`. This lets the Hub pick a non-xet transfer where possible, and a repo that *hard-requires* xet then surfaces the error naturally at the upload seam (defensive `#[cfg(not(feature = "xet"))]` guard on the `xet` branch). This steers the protocol rather than only reacting to it.

### Rejected alternative (considered)

**Plain-HTTP fallback for downloads.** Because the Hub's `resolve` URL 302-redirects to a CAS blob that serves the xet-backed bytes over ordinary HTTP, a xet-off build *could* transparently download xet-backed files without xet. We **rejected** this: silent fallback hides that the fast path is unavailable, produces surprising performance cliffs, and diverges from the "opt out explicitly" contract. Non-xet-backed files still download over plain HTTP as before; only genuinely xet-backed transfers error.

> Observation (pre-existing, out of scope): the `local_dir` download HEAD uses the redirect-following client, so in practice it resolves to the CAS response and doesn't observe `x-xet-hash` — the cache path uses the no-redirect client and detects xet reliably. The gate is in place on both paths regardless; aligning the `local_dir` HEAD with the cache path can be a separate change.

## Feature default: on by default (recommended)

Shipped **`default = ["xet"]`**. Tradeoff:

- **default-ON (chosen):** Non-breaking for existing 1.0 users — upgrading is a no-op. sglang and dynamo already use `default-features = false`, so they are unblocked automatically without any change on their side. New consumers keep the fast path by default.
- **default-OFF (rejected):** Would maximally advance the unblocking goal, but silently drops xet for every existing user on upgrade — a surprising behavioral regression for a point release.

Consumers who want to avoid `hf-xet` (and its tokio floor) opt out:

```toml
hf-hub = { version = "1", default-features = false }
```

### tokio features made explicit

Non-xet code (`tokio::time::sleep` in retry, `tokio::try_join!` in snapshot download, `spawn_blocking` in the upload/cache paths) previously relied on tokio runtime features leaking in transitively via `hf-xet`. Those are now declared directly on hf-hub's tokio dependency (`time`, `macros`, and native-only `rt`) so the crate builds cleanly with `xet` off.

## Test matrix (all green)

Build / clippy `-D warnings`:

- `cargo check/clippy -p hf-hub` (default, xet on)
- `--no-default-features` (xet off)
- `--no-default-features --features xet`
- `--all-features`
- `--no-default-features --features blocking` (blocking without xet)
- `--features blocking` (default + blocking)
- `cargo clippy --workspace --all-targets` and `--all-features`

Tests:

- `cargo test -p hf-hub` (xet on): 190 passed
- `cargo test -p hf-hub --no-default-features` (xet off): 180 passed (the 10 xet-only unit tests are feature-gated)

wasm (`./scripts/verify_wasm.sh`, now checks both configs):

- `-p hf-hub --target wasm32 --no-default-features` (xet off)
- `-p hf-hub --target wasm32 --no-default-features --features xet` (xet on)
- `wasm/smoke` and `wasm/tests`

fmt: `cargo +nightly fmt --check` clean.

### Live verification (HF_TOKEN present)

Downloaded a genuinely xet-backed file (`sentence-transformers/all-MiniLM-L6-v2/model.safetensors`, ~90 MB, `x-xet-hash` present) via the cache path:

- **xet on:** downloads successfully (90,868,376 bytes) — no regression.
- **xet off:** returns `HFError::XetFeatureDisabled { operation: Download }` — error fires, no silent HTTP fallback.

## CI

- Added `--no-default-features` and `--no-default-features --features blocking` clippy steps so the xet-off configuration stays green.
- `verify_wasm.sh` now checks both the xet-off and xet-on wasm builds explicitly.

## Docs

Feature tables updated in `AGENTS.md`, crate docs (`lib.rs` "Cargo features"), and `README.md`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,6 +53,12 @@ jobs:
       - name: Clippy (all features)
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
+      - name: Clippy (no default features — xet disabled)
+        run: cargo clippy -p hf-hub --all-targets --no-default-features -- -D warnings
+
+      - name: Clippy (blocking without xet)
+        run: cargo clippy -p hf-hub --all-targets --no-default-features --features blocking -- -D warnings
+
   build-and-test:
     name: Build & Test
     runs-on: ${{ matrix.runs-on }}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,9 +80,12 @@ browser tests and runtime build flags live under `wasm/`.
 
 | Feature | Default? | Notes |
 |---------|----------|-------|
+| `xet` | **on** | Enables Xet high-performance transfers via the optional `hf-xet` dependency. Gates `src/xet.rs`, `HFClient::xet_session`/`xet_state`, and every xet call site. With it off, xet-backed transfers fail with `HFError::XetFeatureDisabled` (no silent HTTP fallback), the `hf-xet` dep is dropped, and upload negotiation advertises only `basic`/`multipart`. Consumers that need to avoid `hf-xet` (e.g. to escape its `tokio` floor) build with `default-features = false`. |
 | `blocking` | off | Synchronous `*Sync` wrappers in `blocking.rs`. Pulls in `tokio/rt`. Native only. |
 | `rustls-tls` | off | Force the rustls TLS backend on reqwest's native build. No effect on wasm. |
-| `default` | (empty) | No features by default. |
+| `default` | `["xet"]` | Xet on by default so upgrading from 1.0 is non-breaking. |
+
+When adding code that dispatches into `src/xet.rs`, gate the seam with `#[cfg(feature = "xet")]` and add a `#[cfg(not(feature = "xet"))]` arm returning `HFError::xet_feature_disabled(XetOperation::…)`. Any tokio runtime feature relied on by non-xet code must be declared directly in `hf-hub/Cargo.toml` (it must not be assumed to leak in transitively via `hf-xet`).
 
 ### Method builders (bon)
 
@@ -200,9 +203,10 @@ hf-hub/
 │   │   │                           #   runtime/hardware/secrets/variables/duplicate
 │   │   ├── users.rs                # Users component: User/Organization/OrgMembership, whoami,
 │   │   │                           #   user+org lookup, followers/following
-│   │   ├── xet.rs                  # Xet component (pub(crate)): XetConnectionInfo + xet transfer
-│   │   │                           #   plumbing. Most methods are native-only (cached session, fs);
-│   │   │                           #   the wasm `xet_download_stream` builds a fresh session per call
+│   │   ├── xet.rs                  # Xet component (pub(crate), gated on `feature = "xet"`):
+│   │   │                           #   XetConnectionInfo + xet transfer plumbing. Most methods are
+│   │   │                           #   native-only (cached session, fs); the wasm `xet_download_stream`
+│   │   │                           #   builds a fresh session per call
 │   │   ├── buckets/
 │   │   │   ├── mod.rs              # HFBucket handle, bucket types, create/list/tree/batch/download
 │   │   │   └── sync.rs             # BucketSync* types, HFBucket::sync — plan computation and execution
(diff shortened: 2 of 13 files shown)


assafvayner:
## Implementation plan

### Goal
Make `hf-xet` an optional dependency behind a default-on `xet` feature; xet-requiring transfers fail with a clear error when the feature is off (no silent HTTP fallback).

### Steps
1. **Cargo.toml** — `hf-xet` → `optional = true`; `default = ["xet"]`, add `xet = ["dep:hf-xet"]`. Declare the tokio features that non-xet code needs directly (`time`, `macros`, native-only `rt`) since they previously leaked in via `hf-xet`.
2. **error.rs** — add `HFError::XetFeatureDisabled { operation: XetOperation }` (+ `xet_feature_disabled` constructor). `XetOperation` stays ungated so call sites reference it in either config.
3. **lib.rs** — `#[cfg(feature = "xet")] pub(crate) mod xet;` + "Cargo features" doc.
4. **client.rs** — gate `xet_state` field/init, `xet_session`, `replace_xet_session`, and the 7 xet unit tests.
5. **Call sites** — gate each seam with `#[cfg(feature = "xet")]` + a `#[cfg(not(feature = "xet"))]` arm returning `xet_feature_disabled(...)`:
   - `download.rs`: stream, local_dir, cache, and both snapshot batch futures.
   - `upload.rs`: advertise only `basic`/`multipart` when off; defensive error on the `xet` upload branch.
   - `buckets/mod.rs`: `download_file_stream` (partial), `upload_files`/`download_files`/`upload_sources_impl` (whole-body).
   - `buckets/sync.rs`: `execute_download_plan` xet section (delete logic stays unconditional).
6. **CI / wasm** — add xet-off clippy steps; `verify_wasm.sh` checks both xet-on and xet-off wasm builds.
7. **Docs** — feature tables in AGENTS.md, lib.rs, README.

### Testing strategy
- Build + clippy `-D warnings` across the full feature matrix (default / no-default / xet / all-features / blocking combos) and `--workspace`.
- `cargo test -p hf-hub` in both feature states (xet unit tests feature-gated).
- `./scripts/verify_wasm.sh` (both configs) + `wasm/smoke` + `wasm/tests`.
- `cargo +nightly fmt --check`; docs build with `-D warnings --cfg docsrs`.
- Live: download a xet-backed safetensors from `sentence-transformers/all-MiniLM-L6-v2` — succeeds with xet on, returns `XetFeatureDisabled` with xet off.

PR #182, opened 2026-07-03
fix: run blocking API on a dedicated runtime thread so Sync handles are safe in async contexts

assafvayner:
## Problem

The 1.0 `*Sync` handles each own a current-thread `tokio::runtime::Runtime`, and every wrapper method is `self.runtime.block_on(...)` (~65 sites across `blocking.rs` and the sync mirrors in `repository/*.rs`, `buckets/*.rs`, `spaces.rs`, `users.rs`, `cache/mod.rs`). Tokio's enter-guard makes `Runtime::block_on` panic when the calling thread is already inside a runtime, and dropping the runtime from async context panics too. So from async code:

- constructing an `HFClientSync` works,
- any method call panics ("Cannot start a runtime from within a runtime"),
- and even dropping the client panics during unwind.

This is a migration trap. hf-hub 0.4's sync API was ureq-backed (no runtime), so calling it inside async handlers was rude but worked. A mechanical ureq-to-blocking migration keeps compiling, passes plain-thread tests, then panics in production only on paths that run inside a runtime. Consumer validation found real exposure: mistral.rs (sync API threaded through ~22 files, used from the async mistralrs-server), fastembed (sync lib consumed by async frameworks rig/swiftide), and the sglang router (`HFClientSync` in an async server). It also violates our own AGENTS.md rule: never panic.

## Fix: the `reqwest::blocking` model

The runtime moves onto a dedicated background thread owned by the client. Blocking calls from async contexts then block a worker thread — the same rudeness as 0.4's ureq — but they work, and both panic modes disappear.

**Thread ownership.** A private `RuntimeThread` (in `blocking.rs`) spawns a `hf-hub-blocking-runtime` thread that builds a current-thread runtime, sends its `Handle` back, and parks in `runtime.block_on(shutdown_rx)` so foreign threads can spawn onto it. The runtime is created on — and never leaves — that thread, so no caller thread ever enters or drops it (that kills the Drop panic).

**One helper, no per-site boilerplate.** All wrapper methods go through `RuntimeThread::run_future`: they clone their inner `Arc`'d handle into a `'static + Send` future, `handle.spawn` it, and park on a `std::sync::mpsc` channel for the result. A std channel deliberately — tokio's `blocking_recv` has its own enter-guard panic. Stream-collecting methods (`list_*` etc.) run the whole pin-and-drain collection inside the spawned future.

**Panic propagation.** The spawned task is wrapped in `catch_unwind` and the payload is resumed on the caller (`resume_unwind`), matching the semantics `Runtime::block_on` had for panicking tasks. If the runtime thread is ever gone, `run_future` returns `HFError::Other` instead of panicking.

**Clone/shutdown semantics.** `*Sync` handles clone freely and share one background thread (`Arc<RuntimeThread>`). When the last handle drops, the shutdown oneshot sender drops, the receiver resolves, `block_on` returns, and the runtime is dropped on its own thread. The thread is detached rather than joined, so dropping the last handle never blocks the caller (reqwest-style; documented on `RuntimeThread`).

**Public API unchanged.** All bon builder signatures, `HFClientSync::new()`/`from_inner()` (thread-spawn failure maps onto the existing `HFResult` error path), and the xet doc note (xet still spins up its own pool, separate from this runtime — wording updated) are intact. The AGENTS.md blocking-counterpart template now shows the `run_future` pattern.

One incidental change in `repository/download.rs`: two lazy borrowing-closure streams in `snapshot_download` (`head_futs` and `download_concurrently`) are collected eagerly into `Vec`s. Leaving them lazy makes the enclosing future fail the `Send + 'static` bound with rustc's "implementation of `FnOnce` is not general enough" HRTB limitation. (#180's boxing addresses the same spot; if #180 merges first this hunk may partially collapse in the rebase.)

## Alternatives considered

- **Doc-only** ("don't call the sync API from async"): keeps the trap armed; the failing consumers above show it gets hit in practice.
- **`Handle::try_current()` guard returning a typed error**: rejected in favor of making the call work; also false-positives on `tokio::task::block_in_place` blocks (multi-thread runtimes), where `try_current` succeeds but `block_on` would have been legal, so the guard can't be made precise.

## Tests

New regression tests (all panicked before this change, verified by running the same shape against the base commit — panics at the `block_on` site in `repository/mod.rs`):

- unit (`blocking.rs`, network-free via a connection-refused local endpoint): sync `info()` call inside `#[tokio::test]` multi-thread and current-thread runtimes; construct-and-drop of `HFClientSync` inside async context; clones share the runtime thread (`Arc::ptr_eq`) and survive staggered drops; `run_future` resumes task panics on the caller.
- integration (`blocking_test.rs`, HF_TOKEN-gated): `info()` + small `download_file()` from inside a multi-thread `#[tokio::test]` (the exact consumer shape that used to panic), a current-thread-runtime call, and clone/staggered-drop in async context.

Verification:

- `cargo test -p hf-hub --features blocking` / `--all-features`: 198 passed
- `cargo test -p integration-tests --test blocking_test` (live, HF_TOKEN): 25 passed, including the 3 new async-context tests
- `cargo clippy -p hf-hub --all-features -- -D warnings`: clean
- `cargo +nightly fmt --check`: clean
- `./scripts/verify_wasm.sh`: green (blocking stays native-only)

## Stacking

Based on `assaf/wasm-xet-deps`, same as #178/#179/#180/#181, and touches the same sync-mirror files — expect conflicts with all four. Suggest merging this after #180 (its future-boxing overlaps with the `snapshot_download` hunk here and the rebase is easiest in that order).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,7 +80,7 @@ browser tests and runtime build flags live under `wasm/`.
 
 | Feature | Default? | Notes |
 |---------|----------|-------|
-| `blocking` | off | Synchronous `*Sync` wrappers in `blocking.rs`. Pulls in `tokio/rt`. Native only. |
+| `blocking` | off | Synchronous `*Sync` wrappers in `blocking.rs`. Pulls in `tokio/rt` + `tokio/sync`. Native only. |
 | `rustls-tls` | off | Force the rustls TLS backend on reqwest's native build. No effect on wasm. |
 | `default` | (empty) | No features by default. |
 
@@ -124,7 +124,9 @@ helper that already has the field as an `Option<T>` local).
 
 Every async method MUST have a manually-written blocking counterpart on the corresponding
 `*Sync` struct under `#[cfg(feature = "blocking")]`. Mirror the same `#[builder]` parameter list
-verbatim. The body calls the inner async builder and blocks on the runtime:
+verbatim. The body clones the inner handle into a `'static + Send` future and hands it to the
+client's dedicated runtime thread via `run_future` — never call `Runtime::block_on` on a caller
+thread, since that panics when the caller is already inside a tokio runtime:
 
 ```rust
 #[cfg(feature = "blocking")]
@@ -136,21 +138,16 @@ impl crate::blocking::HFRepositorySync {
         #[builder(into)] name: String,
         revision: Option<String>,
     ) -> HFResult<MyOutput> {
-        self.runtime.block_on(
-            self.inner
-                .my_method()
-                .name(name)
-                .maybe_revision(revision)
-                .send(),
-        )
+        let inner = self.inner.clone();
+        self.runtime
+            .run_future(async move { inner.my_method().name(name).maybe_revision(revision).send().await })
     }
 }
 ```
 
-For stream-returning methods, the sync counterpart collects into `Vec<T>`: pin the stream and
-drain with `while let Some` (see existing examples in `repository/listing.rs` and
-`repository/mod.rs`). On `HFSpaceSync`, the runtime is reached via `&self.repo_sync.runtime`
-because the struct has no direct `runtime` field. Do NOT reintroduce the legacy `sync_api!` /
+For stream-returning methods, the sync counterpart collects into `Vec<T>` inside the future
+handed to `run_future`: pin the stream and drain with `while let Some` (see existing examples in
+`repository/listing.rs` and `repository/mod.rs`). Do NOT reintroduce the legacy `sync_api!` /
 `sync_api_stream!` / `sync_api_async_stream!` macros — they were removed deliberately when the
 crate moved to bon.
 
--- a/hf-hub/Cargo.toml
+++ b/hf-hub/Cargo.toml
@@ -34,7 +34,7 @@ sha2 = "0.11"
 
 [features]
 default = []
-blocking = ["tokio/rt"]
+blocking = ["tokio/rt", "tokio/sync"]
 rustls-tls = ["reqwest/rustls"]
 
 [target.'cfg(target_family = "wasm")'.dependencies]
--- a/hf-hub/src/blocking.rs
+++ b/hf-hub/src/blocking.rs
@@ -1,35 +1,122 @@
+use std::future::Future;
+use std::panic::AssertUnwindSafe;
 use std::sync::Arc;
+
+use futures::FutureExt;
 
 use crate::client::HFClient;
 use crate::error::{HFError, HFResult};
 use crate::repository::{HFRepository, RepoType, RepoTypeDataset, RepoTypeKernel, RepoTypeModel, RepoTypeSpace};
 
-fn build_runtime() -> HFResult<Arc<tokio::runtime::Runtime>> {
-    tokio::runtime::Builder::new_current_thread()
-        .enable_all()
-        .build()
-        .map(Arc::new)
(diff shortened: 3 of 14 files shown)


assafvayner:
## Implementation plan

1. **Core primitive** (`hf-hub/src/blocking.rs`): add a private `RuntimeThread` struct.
   - `spawn()` starts a named background thread (`hf-hub-blocking-runtime`) that builds a current-thread runtime, sends its `Handle` back over a std mpsc channel (build failure travels the same channel and maps to the existing `HFError::Other` runtime-creation error), then parks in `runtime.block_on(shutdown_rx)` so the runtime keeps driving and foreign threads can spawn onto it.
   - `run_future(future)` — the single helper every wrapper uses: `handle.spawn` the `'static + Send` future wrapped in `catch_unwind`, park on `std::sync::mpsc::sync_channel(1)::recv` (std, not tokio's `blocking_recv`, which has its own enter-guard panic), `resume_unwind` task panics on the caller, map a dead runtime thread to `HFError::Other`.
   - Shutdown = drop of the last `Arc<RuntimeThread>`: the oneshot sender drops, the thread's `block_on` returns, the runtime is dropped on its own thread; the thread is detached so Drop never blocks the caller.
2. **Struct swap**: `HFClientSync` / `HFRepositorySync<T>` / `HFBucketSync` replace `runtime: Arc<tokio::runtime::Runtime>` with `Arc<RuntimeThread>`; constructors call `RuntimeThread::spawn()?`; signatures unchanged.
3. **Mechanical rewrite of all 65 wrapper sites** (repository/{mod,commits,listing,download,upload}.rs, buckets/{mod,sync}.rs, spaces.rs, users.rs, cache/mod.rs): `self.runtime.block_on(EXPR)` becomes `let inner = self.inner.clone(); self.runtime.run_future(async move { EXPR'.await })`, with stream-collecting bodies moved wholesale into the spawned future. Done with a script + `cargo +nightly fmt`, then hand-fixed the two `snapshot_download` HRTB failures by eagerly collecting the borrowing-closure streams.
4. **Feature/deps**: `blocking = ["tokio/rt", "tokio/sync"]` (oneshot for shutdown).
5. **Docs**: reword the xet note on `HFClientSync` ("runtime on the background thread owned by `HFClientSync`"), update the AGENTS.md blocking-counterpart template to the `run_future` pattern, drop the stale `HFSpaceSync` sentence, update the feature table.
6. **Tests**:
   - Network-free unit tests in `blocking.rs` using a connection-refused local endpoint: calls from `#[tokio::test]` (both flavors), construct+drop in async context, clone sharing/staggered drops, panic-propagation through `run_future`.
   - HF_TOKEN-gated integration tests in `blocking_test.rs`: `info()` + `download_file()` inside a multi-thread runtime, current-thread-runtime call, clone/staggered drop.
   - Confirmed the regression by running the new test shape against the base commit: panics at the `block_on` site (`repository/mod.rs:1668`).
7. **Verification**: `cargo test -p hf-hub --features blocking` and `--all-features` (198 passed), live `blocking_test` suite (25 passed), clippy `-D warnings` clean, fmt clean, `./scripts/verify_wasm.sh` green, `cargo check --workspace --all-features` (hfrs, examples, benches) green.

assafvayner:
Superseded by #186, which keeps the same dedicated-runtime-thread execution model but drops the `'static` bound via a scoped-thread `Handle::block_on`, so all ~65 wrapper call sites stay unchanged (4 files instead of 14).

PR #185, opened 2026-07-07
feat: allow passing params by ref

### LINKED ISSUE #184: Accept `&str` or `impl ToOwned`
https://github.com/huggingface/kernels/pull/694#discussion_r3534538412

Cc: @danieldk

McPatate:
In a lot of different places, `String` is not necessary. Pass by ref or make it `#[builder(into)] var: String` so that you can call `.to_string` on `var`. This accepts `&str` too.

closes #184

### DIFF
--- a/examples/blocking_spaces.rs
+++ b/examples/blocking_spaces.rs
@@ -26,7 +26,7 @@ fn main() -> hf_hub::HFResult<()> {
     client
         .create_repository()
         .repo_type(RepoTypeSpace)
-        .repo_id(space.repo_path())
+        .repo_id(&space.repo_path())
         .private(true)
         .space_sdk("static")
         .exist_ok(true)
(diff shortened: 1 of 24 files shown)


examples/blocking_spaces.rs
```diff
@@ -26,7 +26,7 @@ fn main() -> hf_hub::HFResult<()> {
     client
         .create_repository()
         .repo_type(RepoTypeSpace)
-        .repo_id(space.repo_path())
+        .repo_id(&space.repo_path())
```
assafvayner, line 29: could repo_path() return a &str?

assafvayner (approved):
minor nit, non-blocker

examples/blocking_spaces.rs
McPatate: There's `format!` in one of the return paths, I think it makes sense to return a `String` here

PR #188, opened 2026-07-09
fix(ci): publish release to crates.io explicitly

assafvayner:
## Problem

The `Rust Release` workflow failed on the `v1.0.0-rc.2` tag push ([run](https://github.com/huggingface/hf-hub/actions/runs/29033772628/job/86173174909)):

```
error: crates-io is replaced with non-remote-registry source registry `hf-mirror`;
include `--registry crates-io` to use crates.io
```

The `setup-hf-registry-proxies` action rewrites the `crates-io` source to point at the internal `hf-mirror` proxy (great for *fetching* deps behind the proxy). But that mirror is not a remote registry you can *publish* to, so a bare `cargo publish` is redirected there and errors out.

## Fix

Pass `--registry crates-io` to `cargo publish` so the publish targets crates.io directly, bypassing the source replacement. This is exactly what cargo's error message recommends.

## Testing

- CI (this workflow) only runs on tag pushes, so it can't be exercised by this PR directly.
- Once merged, re-pushing the `v1.0.0-rc.2` tag (or the next release tag) should publish successfully. The `--token` flag continues to authenticate against crates.io as before.

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -28,10 +28,8 @@ jobs:
 
       - name: Publish package rust
         working-directory: ./hf-hub
-        # The HF registry proxy setup replaces the crates-io source with a
-        # non-remote-registry mirror (hf-mirror), which cannot be published to.
-        # Pass --registry crates-io so the publish targets crates.io directly.
-        # The token is provided via CARGO_REGISTRY_TOKEN rather than the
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-release.yml
McPatate, line 1: looks good but I think we should setup the trusted publisher thing instead of relying on custom tokens

McPatate (approved):

PR #190, opened 2026-07-28
Match Python cache fallback when symlinks are unavailable

assafvayner:
## Summary

Follow-up to #172.

- try Windows file symlinks before falling back to copied snapshot files
- move freshly downloaded blobs into snapshots when symlinks are unavailable, while preserving shared blobs by copying them
- reuse regular snapshot files from degraded caches
- update named revision refs when a cache snapshot is reused, preserving subsequent offline cache resolution

## Plan

1. Align degraded no-symlink cache handling with Python.
2. Preserve revision-ref updates when reusing existing snapshots.
3. Cover degraded-cache reuse and offline lookup with regression tests.

## Validation

- `cargo +nightly fmt`
- `cargo test -p hf-hub`


assafvayner:
Implementation plan and verification:

1. Align the no-symlink cache fallback with Python’s move-for-new / copy-for-existing behavior.
2. Keep named revision refs current when an existing snapshot is reused.
3. Cover degraded-cache reuse followed by an offline lookup.

Validated with `cargo +nightly fmt`, `cargo test -p hf-hub`, and `cargo clippy -p hf-hub --all-features -- -D warnings`.

assafvayner:
Implementation plan for the cache correctness follow-up:

1. Prefer `X-Repo-Commit` from 304 responses so named refs advance when a branch moves without changing the file ETag.
2. Keep blob classification, download, and snapshot-pointer creation under the same per-ETag cache lock for HTTP, single-file Xet, and batched Xet paths.
3. Add deterministic regressions for 304 ref advancement and no-symlink concurrent pointer materialization.
4. Validate with `cargo +nightly fmt`, `cargo test -p hf-hub`, `cargo clippy -p hf-hub --all-features -- -D warnings`, and `./scripts/verify_wasm.sh`.

McPatate (approved):

PR #191, opened 2026-07-28
fix: don't trust redirect Content-Length as file size

i386:
## Summary

Prevent redirect response bodies from being mistaken for the size of the file they point to.

`extract_file_size` previously fell back to `Content-Length` whenever `X-Linked-Size` was absent. For a `3xx` response from the Hub resolve endpoint, that value describes the redirect body rather than the target file. The download path could therefore record and propagate a small, incorrect expected size.

This keeps `X-Linked-Size` authoritative on every response, but only uses `Content-Length` for non-redirect responses. A redirect without `X-Linked-Size` now leaves the size unknown instead of propagating an incorrect value into metadata, progress totals, or Xet size hints.

The change is rebased from Mesh-LLM/hf-hub@27573dc6036a273992a4334d9eb371ecd7359e16 and preserves the original author.

## Python implementation

Python's `huggingface_hub` uses the same metadata request strategy proposed in #176:

- send a `HEAD` request with `Accept-Encoding: identity`
- follow relative redirects, such as repository renames
- stop at absolute redirects to external storage or a CDN

Its size extraction currently selects:

```python
response.headers.get("X-Linked-Size") or response.headers.get("Content-Length")
```

That means Python has the same edge case when an absolute redirect does not include `X-Linked-Size`: `Content-Length` is the size of the redirect response, not the target file. Python later requires the size to be present and raises `FileMetadataError` if it is unknown.

Rust already represents the extracted size as `Option<u64>` and tolerates an unknown size in its download paths. This PR intentionally uses that representation rather than accepting a known-wrong value. It is complementary to #176: that PR follows relative redirects to reach their metadata, while this change handles redirects that intentionally remain unfollowed.

## Tests

Added focused unit coverage for:

- preferring `X-Linked-Size` over `Content-Length`
- using `Content-Length` on a normal response
- ignoring `Content-Length` on a redirect
- trusting `X-Linked-Size` on a redirect
- preserving the existing behavior for an invalid `X-Linked-Size`

Validation:

- `cargo test -p hf-hub` — 189 unit tests, 2 integration-style future-size tests, and 22 doctests passed
- nightly `cargo fmt -- --check`
- `./scripts/verify_wasm.sh`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1106,6 +1106,7 @@ dependencies = [
  "getrandom 0.2.17",
  "globset",
  "hf-xet",
+ "http",
  "hyper",
  "pathdiff",
  "percent-encoding",
--- a/hf-hub/Cargo.toml
+++ b/hf-hub/Cargo.toml
@@ -49,4 +49,5 @@ wasm-bindgen-futures = "0.4"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 tempfile = "3"
 serial_test = "3"
+http = "1"
 
--- a/hf-hub/src/repository/files.rs
+++ b/hf-hub/src/repository/files.rs
@@ -352,9 +352,17 @@ pub(super) fn extract_commit_hash(response: &reqwest::Response) -> Option<String
 
 pub(crate) fn extract_file_size(response: &reqwest::Response) -> Option<u64> {
     let headers = response.headers();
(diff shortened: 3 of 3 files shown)


assafvayner (approved):

assafvayner:
test failures expected for outside contributor PRs, merging.

PR #192, opened 2026-07-29
Upgrade hf-xet to 1.5.4 to pick up a fix for download hangs

seanses:
- Specify `hf-xet` at minor instead of patch verison.
- Update Cargo.lock by running `cargo update -p hf-xet`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,15 +176,6 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.10.4"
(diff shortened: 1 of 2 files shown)


assafvayner:
setting to 1.5 doesn't enforce late enough version, 1.5.3 is the same as 1.5.3 and any higher minor/patch version, for 1.5.4 hf-xet version we will need to set 1.5.4 (1.5 only still allows `1.5.*`

assafvayner:
closing in favor of #197

PR #194, opened 2026-08-01
fix: require tower-http >= 0.6.9 so bracketed-filename redirects are followed

assafvayner:
## Problem

Downloading any file whose path contains `[` or `]` fails with a bare `HTTP error: 302 Found` instead of returning the file.

```
hf-hub error: HTTP error: 302 Found
  (url=https://huggingface.co/datasets/dsixteen/processed/resolve/<rev>/final/suiton00/[suiton00]%20Junpei%20x%20Ruo%20%231/99595347_p1.png,
   request_id=Root=1-6a6dbeee-6f455af9287fdb5e78bd4bf9)
```

Note the literal brackets — that's the bug.

## Cause

`url::Url` implements the WHATWG URL Standard, which leaves `[`, `]`, `^`, and `|` unencoded in paths; RFC 3986 forbids them.

reqwest delegates redirect handling to `tower-http`'s `FollowRedirect`. Before tower-http 0.6.9, it resolved `Location` against the request URI using `iri-string`'s strict RFC 3986 parser:

```rust
fn resolve_uri(relative: &str, base: &Uri) -> Option<Uri> {
    let relative = UriReferenceStr::new(relative).ok()?;
    let base = UriAbsoluteString::try_from(base.to_string()).ok()?;   // <-- fails on raw [ ]
    Uri::try_from(relative.resolve_against(&base).to_string()).ok()
}
```

When that returns `None` the middleware returns the 3xx unfollowed rather than erroring, so `check_response` sees a non-2xx and maps it to `HFError::Http`.

## Fix

tower-http 0.6.9 ([tower-rs/tower-http, May 2026](https://github.com/tower-rs/tower-http)) dropped `iri-string` from `follow_redirect` and resolves `Location` with `url::Url::join` instead — the same WHATWG parser hf-hub builds URLs with, so they agree by construction:

```rust
fn resolve_uri(relative: &str, base: &Uri) -> Option<Uri> {
    let base_url = Url::parse(&base.to_string()).ok()?;
    let resolved = base_url.join(relative).ok()?;
    Uri::try_from(String::from(resolved)).ok()
}
```

reqwest (including latest 0.13.4) still declares `tower-http = "^0.6.8"`, so nothing in the dependency graph forces the fixed version — a consumer with a stale lockfile keeps the bug. This PR adds a direct `tower-http = { version = "0.6.9", default-features = false }` dependency (non-wasm only, where reqwest uses it) purely to raise the floor for downstream lockfiles, and updates our own lockfiles to 0.6.11.

An earlier revision of this PR instead percent-encoded all path components to strict RFC 3986 in `client.rs`. That remains the "fix at the source" option, but the dependency floor resolves the observed failure with no behavior change to URL construction. Known trade-offs accepted here:

- Request URIs still carry raw `[ ] ^ |` on the wire (WHATWG-legal, RFC-illegal). The Hub and its CDN accept them today.
- A `\` in a repo id is still rewritten to `/` by `Url::parse` (latent — Hub naming rules forbid it).

## Tests

`redirect_is_followed_for_a_filename_with_brackets` — end-to-end against a local server that 302s; reproduces the original failure and guards the version floor. Confirmed to fail against tower-http 0.6.8 with this branch's URL-building code, and pass at 0.6.11:

```
tower-http 0.6.8:  redirect was not followed, landed on .../[artist]%20name.png (302)
tower-http 0.6.11: passes
```

```
cargo test -p hf-hub --all-features            218 passed
cargo clippy -p hf-hub --all-features --all-targets -- -D warnings   clean
cargo +nightly fmt --check                     clean
./scripts/verify_wasm.sh                       green
```

## Impact

Found while investigating a production scan-error-rate alert on repository-scanner. Bracketed filenames are common in image and archive datasets that group files into `[artist]`-style directories; one such dataset failed ~230k blob downloads in a single night, pushing that service's scan error rate from a ~2% baseline to 60%.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1119,6 +1119,7 @@ dependencies = [
  "tokio",
  "tokio-retry",
  "tokio-util",
+ "tower-http",
  "tracing",
  "url",
  "wasm-bindgen-futures",
@@ -1477,16 +1478,6 @@ name = "ipnet"
 version = "2.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d98f6fed1fde3f8c21bc40a1abb88dd75e67924f9cffc3ef95607bad8017f8e2"
-
-[[package]]
-name = "iri-string"
-version = "0.7.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c91338f0783edbd6195decb37bae672fd3b165faffb89bf7b9e6942f8b1a731a"
-dependencies = [
- "memchr",
- "serde",
-]
 
 [[package]]
 name = "is-terminal"
@@ -2954,20 +2945,20 @@ dependencies = [
 
 [[package]]
 name = "tower-http"
-version = "0.6.8"
+version = "0.6.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d4e6559d53cc268e5031cd8429d05415bc4cb4aefc4aa5d6cc35fbf5b924a1f8"
+checksum = "4cfcf7e2740e6fc6d4d688b4ef00650406bb94adf4731e43c096c3a19fe40840"
 dependencies = [
  "bitflags",
  "bytes",
  "futures-util",
  "http",
  "http-body",
- "iri-string",
  "pin-project-lite",
  "tower",
  "tower-layer",
  "tower-service",
+ "url",
 ]
 
 [[package]]
--- a/hf-hub/Cargo.toml
+++ b/hf-hub/Cargo.toml
@@ -39,6 +39,15 @@ blocking = ["tokio/rt"]
 rustls-tls = ["reqwest/rustls"]
 socks = ["reqwest/socks"]
 
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+# Not used directly: raises the floor on reqwest's redirect middleware. Before
+# 0.6.9, tower-http resolved `Location` against the request URI with a strict
+# RFC 3986 parser and silently returned the 3xx unfollowed when the request
+# path contained characters the WHATWG `url` crate leaves unencoded but RFC
+# 3986 forbids (`[`, `]`, `^`, `|`), so downloads of such filenames failed
+# with a bare `302 Found`. reqwest itself only requires 0.6.8.
+tower-http = { version = "0.6.9", default-features = false }
+
 [target.'cfg(target_family = "wasm")'.dependencies]
 # `tokio-retry` -> `rand` 0.8 -> `rand_core` 0.6 -> `getrandom` 0.2 transitively
 # needs the `js` feature on wasm32 to compile.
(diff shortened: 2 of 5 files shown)


assafvayner:
Follow-up on the encode set: rather than trusting a hand-enumerated `pchar` table, the set is now pinned by a test against `iri_string`'s `from_path_segment` over every ASCII character.

I did consider using `iri-string` directly in `append_path_segments` instead of `percent-encoding` — its output is byte-identical on every case I tried, including the `%` handling (`a%2Fb` → `a%252Fb`). I kept it out of the production path because it is **not** in the wasm dependency tree: reqwest's wasm backend goes through the browser's fetch API and never pulls `tower-http`, so making it a normal dependency would add a crate to wasm builds for a redirect problem that cannot occur there. As a dev-dependency it costs nothing on any target — it is already in the native tree via `tower-http`, and `cargo tree -e normal --target wasm32-unknown-unknown` shows it absent.

The differential test is what makes the table trustworthy, and I verified it catches drift: un-encoding `[` and `^` fails it with `PATH_SEGMENT_ENCODE_SET disagrees with iri-string for "[" (0x5b)`.

Happy to switch to using `iri-string` directly in the implementation if you would rather have one encoder than a table plus a test — it is a small change either way.

assafvayner:
Rewrote and force-pushed — the earlier version was too narrow to review sensibly.

It fixed `append_path_segments` only, which treated the bug as "`path_segments_mut().push()` uses the wrong character set." That's not the actual defect. `url::Url` is a WHATWG URL implementation, and *every* entry point into it is equally lax — including `Url::parse`, which is how repo ids, bucket ids, usernames, and organizations were reaching the URL after being interpolated into a `format!` raw:

```
Url::parse("https://huggingface.co/datasets/owner/re[po]/resolve/main")
  → /datasets/owner/re[po]/resolve/main    follows=false
```

So the first version closed one door and left the rest open, while the commit message implied the class was handled. Now every path component goes through one encoder, and the builders (`api_url`, `download_url`, `bucket_api_url`) encode internally so call sites can't forget one.

Also picked up a second symptom along the way: `\` isn't just left unencoded, the WHATWG parser rewrites it to `/`, so `owner\name` silently addresses `owner/name`.

The `iri-string` dev-dependency discussion in my earlier comment still applies unchanged — it stays out of the production tree on all targets, and now backs two guards rather than one: the ASCII-wide character-table pin, and a new `url_builders_emit_valid_rfc_3986_for_hostile_input` that asserts every builder's output parses as `UriAbsoluteString`. I checked both fail against the unencoded implementation rather than passing vacuously.

assafvayner:
Correction to what I wrote earlier in this PR: I claimed the WHATWG path set leaves `\` unencoded. That's true of the set as specified, but `url`'s `path_segments_mut().push()` uses a stricter set that does encode it:

```
push("back\slash")     -> /back%5Cslash      correctly encoded
parse("/back\slash")   -> /back/slash        rewritten to a separator
```

So the `\` hazard is real but belongs to the `Url::parse` door (repo ids, bucket ids), not the `push` door (filenames). Fixed in the doc comment, commit message, and description.

While checking that, I also measured the actual delta rather than describing it. The old `push()` encoded 45 of the 128 ASCII characters — all C0 controls, space, `"`, `#`, `%`, `/`, `<`, `>`, `?`, `` ` ``, `{`, `}`, DEL — plus all non-ASCII. It was short by exactly four: `[`, `]`, `^`, `|`. Worth stating precisely, since "the old code didn't encode things" overstates it; the set was nearly right and missed a specific corner.

McPatate (approved):

PR #198, opened 2026-08-06
Add info() on HFRepository<RepoTypeAny> returning runtime-tagged RepoInfo

assafvayner:
## Summary

Adds `info()` to the runtime-tagged repo handle `HFRepository<RepoTypeAny>`, so callers whose repo kind is only known at runtime (CLI flag, config, upstream enum) can fetch repo metadata without matching over repo kinds.

The method returns a new `RepoInfo` enum wrapping the existing per-kind structs (`ModelInfo`, `DatasetInfo`, `SpaceInfo`, `KernelInfo`):

- **Common accessors** read fields shared by all kinds without a match: `id()`, `author()`, `sha()`, `private()`, `gated()`, `likes()`, `last_modified()`, `internal_id()`, `repo_type()`.
- **Union accessors** cover fields most kinds have, returning `None` where a kind lacks them (documented per method): `downloads()` (Spaces), `created_at()`, `tags()`, `card_data()`, `siblings()`, `trending_score()`, `used_storage()` (kernels).
- **Narrowing** via `as_*()` / `into_*()` or a plain `match` yields the exact same per-kind struct the typed handles return — no field duplication.

The dispatch match lives once inside the library; `fetch_repo_info` already builds the correct URL from the runtime variant, so only the deserialization target differs per kind.

## Changes

- `RepoInfo` enum + accessors in `repository/mod.rs`. `#[non_exhaustive]`; the `Model` variant is boxed (clippy `large_enum_variant`: `ModelInfo` ~896B vs 648B next-largest), following the `HFError`/`HttpErrorContext` precedent. Field access auto-derefs so match ergonomics are unaffected.
- `HFRepository<RepoTypeAny>::info()` bon builder (`HFAnyRepoInfoBuilder`), mirroring the four per-kind `info` impls, plus the mandatory blocking counterpart `HFRepositorySync<RepoTypeAny>::info()` (`HFAnyRepoInfoSyncBuilder`).
- `RepoTypeAny` rustdoc updated — it previously stated `info` was not available on the any-handle.
- `examples/repo_type_any.rs` extended to demonstrate `info()`, common accessors, and variant matching.
- `Cargo.lock`: one-line sync (stale `serial_test` entry from an earlier manifest change); regenerated by any build.

## Testing

- 3 unit tests: common accessors across all four kinds, union-accessor kind gaps, narrowing conversions.
- 2 read-only integration tests (`test_repo_type_any_info`, `test_sync_repo_type_any_info`), run live against the Hub and passing.
- `cargo test -p hf-hub --all-features` (215 passed), `cargo clippy -p hf-hub --all-features -- -D warnings` clean, `cargo +nightly fmt`, `./scripts/verify_wasm.sh` green, examples build.

### DIFF
--- a/hf-hub/src/repository/mod.rs
+++ b/hf-hub/src/repository/mod.rs
@@ -259,27 +259,171 @@ pub struct EvalResultSource {
     pub org: Option<String>,
 }
 
-/// Parsed YAML metadata from a repo card (`README.md` front matter).
-///
-/// The schema varies by repo kind and by the libraries/tags a repo declares, so this wraps the
-/// raw JSON value rather than modeling every possible field. [`Deref`](std::ops::Deref) to
-/// [`serde_json::Value`] for field lookups (e.g. `card_data.get("license")`).
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
-#[serde(transparent)]
-pub struct CardData(pub serde_json::Value);
-
-impl std::ops::Deref for CardData {
-    type Target = serde_json::Value;
-
-    fn deref(&self) -> &Self::Target {
-        &self.0
+// Several repo-card YAML fields (`license`, `language`, `datasets`, ...) accept either a single
+// string or a list of strings; normalize both shapes to `Vec<String>`.
+fn deserialize_string_or_vec<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StringOrVec {
+        One(String),
+        Many(Vec<String>),
     }
+
+    Ok(Option::<StringOrVec>::deserialize(deserializer)?.map(|value| match value {
+        StringOrVec::One(s) => vec![s],
+        StringOrVec::Many(v) => v,
+    }))
 }
 
-impl From<serde_json::Value> for CardData {
-    fn from(value: serde_json::Value) -> Self {
-        Self(value)
-    }
+/// Model Card metadata parsed from a model repo's `README.md` YAML front matter.
+///
+/// Mirrors `huggingface_hub.ModelCardData`. Fields the card declares that aren't modeled above
+/// (e.g. `model-index`, `model_name`, or repo-specific custom keys) are preserved in `extra`.
+#[derive(Debug, Clone, PartialEq, Default, Deserialize, Serialize)]
+pub struct ModelCardData {
+    /// Base model(s) this model derives from (e.g. a fine-tune or adapter).
+    #[serde(default, deserialize_with = "deserialize_string_or_vec")]
+    pub base_model: Option<Vec<String>>,
+    /// Dataset(s) used to train this model.
+    #[serde(default, deserialize_with = "deserialize_string_or_vec")]
+    pub datasets: Option<Vec<String>>,
+    /// Language(s) of the model's training data, as ISO 639 codes or special values like "code".
+    #[serde(default, deserialize_with = "deserialize_string_or_vec")]
+    pub language: Option<Vec<String>>,
+    /// Library used by this model (e.g. `"transformers"`, `"timm"`).
+    #[serde(default)]
+    pub library_name: Option<String>,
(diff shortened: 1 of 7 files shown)


assafvayner:
## Implementation plan

**Design goal:** call the info API on a repo whose kind is only known at runtime, with no per-kind match at the call site; common metadata readable without matching, kind-specific data via a single match on the result.

1. **`RepoInfo` enum** wrapping the existing per-kind info structs (no field duplication; a match yields the same types the typed handles return). Accessor methods for kind-common fields, union accessors returning `None` where a kind lacks the field, `as_*`/`into_*` narrowing helpers.
2. **`HFRepository<RepoTypeAny>::info()`** — additive concrete impl beside the four typed ones (no changes to the sealed `RepoType` trait or existing public builders). Dispatches once internally on the runtime variant; reuses `fetch_repo_info`, which already derives the endpoint from the variant.
3. **Blocking mirror** `HFRepositorySync<RepoTypeAny>::info()` per the crate's mandatory sync-counterpart rule.
4. **Docs/example**: fix the `RepoTypeAny` rustdoc claim that `info` is unavailable; extend `examples/repo_type_any.rs`.

**Testing strategy**

- Unit tests (`repository/mod.rs`): accessors across all four variants built from representative JSON, union-accessor gaps (Space `downloads`, kernel slim shape), narrowing conversions.
- Integration tests: read-only `info()` via `RepoTypeAny` for model + dataset (async) and model (blocking parity), asserting variant tag, `id`, and common accessors against live Hub responses.
- Manual verification: `cargo test -p hf-hub --all-features`, clippy `-D warnings`, `cargo +nightly fmt`, `./scripts/verify_wasm.sh`, example build.

**Alternatives rejected:** associated `type Info` on `RepoType` (breaks the four public named builders; untagged deserialization is unreliable with these overlapping shapes), flat common struct + extras enum (splits existing public types), `RepoInfoCommon` trait (can be layered on later, additively).

hf-hub/src/repository/mod.rs
```diff
@@ -513,6 +513,261 @@ pub struct KernelInfo {
...
+            Self::Model(info) => info.private,
+            Self::Dataset(info) => info.private,
+            Self::Space(info) => info.private,
+            Self::Kernel(info) => info.private,
+        }
+    }
+
+    /// Gated-access state. Either the boolean `false` (open) or the string `"auto"`/`"manual"`
+    /// indicating the approval mode for access requests. Raw JSON because the field is union-typed.
+    pub fn gated(&self) -> Option<&serde_json::Value> {
```
McPatate: not a big fan of having `serde_json::Value` as the return type. We need to either expose a proper type, or a bool imo

hf-hub/src/repository/mod.rs
```diff
@@ -513,6 +513,261 @@ pub struct KernelInfo {
...
+            Self::Model(info) => info.tags.as_deref(),
+            Self::Dataset(info) => info.tags.as_deref(),
+            Self::Space(info) => info.tags.as_deref(),
+            Self::Kernel(_) => None,
+        }
+    }
+
+    /// Parsed repo card (README front matter) data. Always `None` for kernels — the slim
+    /// kernel info shape does not include it.
+    pub fn card_data(&self) -> Option<&serde_json::Value> {
```
McPatate: Similarly, we should wrap in a new type rather than exposing the raw `Value`

hf-hub/src/repository/mod.rs
```diff
@@ -258,6 +259,173 @@ pub struct EvalResultSource {
...
+    pub metrics: Option<Vec<String>>,
+    /// Pipeline tag associated with the model (e.g. `"text-classification"`).
+    #[serde(default)]
+    pub pipeline_tag: Option<String>,
+    /// Hub tags declared on the card.
+    #[serde(default)]
+    pub tags: Option<Vec<String>>,
+    /// Fields present in the card's YAML front matter that aren't modeled above.
+    #[serde(flatten)]
+    pub extra: serde_json::Map<String, serde_json::Value>,
```
McPatate: `HashMap` doesn't work here? Still not a fan of the `Value` but I guess we won't be able to get rid of it easily, so ok to leave it!

McPatate (approved):

PR #199, opened 2026-08-12
fix: adjust create pr request

drbh:
this pr fixes `create_pr` uploads for users who do not have write access to the target repo.

this caused the [`sgl-flash-attn3` upload in kernels-community](https://github.com/huggingface/kernels-community/actions/runs/31569368036/job/94028009394#step:11:57) to fail when the release workflow tried to open a hub pr against the external repo.

the hub stores pr commits under `refs/pr/<number>` in the target repo. previously `hf-hub` requested a pr for the final commit, but still negotiated the lfs and xet uploads against the target branch. this caused the hub to apply a normal branch permission check and reject external contributors before the pr commit could be created.

the pr upload flow now matches the python `huggingface_hub` client:

- pass `create_pr=1` to preupload
- omit `ref` from the lfs batch request
- request the xet write token from `/xet-write-token/None?create_pr=1`
- deserialize `pullRequestUrl` into `CommitInfo::pr_url`

normal branch uploads continue to send their branch ref, xet read and bucket token requests are unchanged, and there are no public api changes.

I've reproduced the issue against [`kernels/drbh/test-repo`](https://huggingface.co/kernels/drbh/test-repo) using a second account without write access. a normal upload and the unpatched `create_pr` flow were both rejected during the lfs batch request. with this patch, the same upload created [pr 1](https://huggingface.co/kernels/drbh/test-repo/discussions/1), left `main` unchanged, and uploaded both the regular and lfs/xet files under `refs/pr/1`. the returned `CommitInfo::pr_url` also contained the new pr url.

the regression tests can be run with

```bash
cargo test -p hf-hub create_pr
```

which outputs

```text
running 2 tests
test repository::files::tests::test_commit_info_with_create_pr_response ... ok
test repository::upload::tests::create_pr_calls_preupload_url ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 183 filtered out
```

please let me know if i should make any changes!

### DIFF
--- a/hf-hub/src/repository/files.rs
+++ b/hf-hub/src/repository/files.rs
@@ -146,7 +146,8 @@ pub struct CommitInfo {
     #[serde(default)]
     pub commit_oid: Option<String>,
     /// Pull-request URL, when `create_pr` was enabled and a PR was opened.
-    #[serde(default)]
+    // the Hub response uses `pullRequestUrl`, matching https://github.com/huggingface/huggingface_hub/blob/ca9284561d5b7049496797992cf01d0bb139ce52/src/huggingface_hub/hf_api.py#L5139.
+    #[serde(default, alias = "pullRequestUrl")]
     pub pr_url: Option<String>,
     /// Pull-request number, when `create_pr` was enabled and a PR was opened.
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


assafvayner (approved):
LGTM

PR #202, opened 2026-09-03
Harden download_file_to_bytes against untrusted Content-Length (CWE-770)

li-jin-quan:
`HFRepository::download_file_to_bytes` (and its `range` variant) pre-allocated its destination `BytesMut` from the server-declared `Content-Length` / `X-Linked-Size` response header, before any body byte is read. That header is not trustworthy against an untrusted endpoint — a malicious mirror, an `HF_ENDPOINT` override, or a MITM — so a hostile server could advertise a multi-gigabyte size while shipping only a handful of bytes and force a large upfront allocation (CWE-770). Grow the buffer from the bytes that actually arrive instead.

- No public API or observable behavior change: the collected `Bytes` are identical.
- `BytesMut` already grows geometrically, so the cost for legitimate responses is a few reallocations.
- The on-disk path (`download_file_to_local_dir` / `snapshot_download`) streams to a file and was never affected; only the in-memory collector trusted the header.


li-jin-quan:
Hi — pinging this after about a week.

download_file_to_bytes pre-allocates using Content-Length from the response. A server that reports a bogus value can make the client allocate a huge buffer up front. This one reads incrementally instead.

Is this something you'd want, or should I close it?

assafvayner:
Hi @li-jin-quan simplifying it a bit, but otherwise fair change

assafvayner (approved):

assafvayner:
CI failures are ok, token is not propagated for external contributor PR's

li-jin-quan:
Thanks for landing this — the `BytesMut::new()` version is simpler than mine, agreed.

And thanks for the note on CI: good to know those red checks are an external-contributor token thing, not my change.

PR #203, opened 2026-09-03
Run cache directory scan on a blocking thread pool

assafvayner:
`scan_cache_dir` walked the whole cache directory synchronously inside an async fn, so on a large cache it monopolized the runtime thread for the duration of the walk. This moves the walk onto `spawn_blocking`; the public API and result are unchanged.

- Split the walk into `scan_cache_dir_blocking` and wrap it in `tokio::task::spawn_blocking`.
- Add a test on a current-thread runtime that fails if the scan runs inline and starves a concurrent timer.

### DIFF
--- a/hf-hub/src/cache/storage.rs
+++ b/hf-hub/src/cache/storage.rs
@@ -231,6 +231,13 @@ fn scan_snapshot(snap_path: &Path, warnings: &mut Vec<String>) -> Vec<CachedFile
 }
 
 pub(crate) async fn scan_cache_dir(cache_dir: &Path) -> crate::error::HFResult<HFCacheInfo> {
+    let cache_dir = cache_dir.to_path_buf();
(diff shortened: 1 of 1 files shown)


hf-hub/src/cache/storage.rs
```diff
@@ -519,6 +526,47 @@ mod tests {
         assert_eq!(result.size_on_disk, 0);
     }
 
+    // Deliberately a default (current-thread) runtime: on a single-threaded executor, an inline
+    // synchronous walk would monopolize the only worker thread and starve the timer below, so
+    // this only passes if the walk actually runs off-thread via `spawn_blocking`.
+    #[tokio::test]
+    async fn test_scan_cache_does_not_block_the_runtime() {
```
McPatate, line 533: the test seems a little overkill imo

PR #205, opened 2026-09-03
Abort the xet download progress poller when its download future is dropped

assafvayner:
The xet download progress poller was spawned as a detached task and only aborted on the happy path after the download future resolved. If the download future was dropped or cancelled first, the poller kept running. This wraps the `JoinHandle` in an `AbortOnDrop` guard so the poller's lifetime is tied to the download call.

- Replace the explicit `abort()` calls with `drop(poll_handle)` on all four download paths.
- Test that dropping the guard tears down the spawned task.

### DIFF
--- a/hf-hub/src/xet.rs
+++ b/hf-hub/src/xet.rs
@@ -126,15 +126,30 @@ fn emit_remaining_completes(progress: &Option<Progress>, tracked: &[TrackedDownl
     }
 }
 
+/// Owns a spawned task's [`JoinHandle`](tokio::task::JoinHandle) and aborts it on drop.
+///
+/// Used to tie a detached poller task's lifetime to a local variable: if the enclosing future
+/// (e.g. a download call) is dropped or aborted before it explicitly tears the poller down, this
+/// guard's `Drop` still runs and cancels the task, so the poller cannot outlive its download.
+#[cfg(not(target_family = "wasm"))]
+struct AbortOnDrop(tokio::task::JoinHandle<()>);
+
+#[cfg(not(target_family = "wasm"))]
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
 #[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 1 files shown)


hf-hub/src/xet.rs
```diff
@@ -618,9 +633,7 @@ impl<T: RepoType> HFRepository<T> {
         let poll_handle = spawn_download_progress_poller(progress, &group, Arc::clone(&tracked));
 
         let result = group.finish().await;
-        if let Some(h) = poll_handle {
-            h.abort();
-        }
+        drop(poll_handle);
```
McPatate, line 636: the `.abort` formulation is a bit more explicit than calling `drop` on it, I'd leave it for readability, or cf my other comment

hf-hub/src/xet.rs
```diff
@@ -126,15 +126,30 @@ fn emit_remaining_completes(progress: &Option<Progress>, tracked: &[TrackedDownl
...
+/// guard's `Drop` still runs and cancels the task, so the poller cannot outlive its download.
+#[cfg(not(target_family = "wasm"))]
+struct AbortOnDrop(tokio::task::JoinHandle<()>);
+
+#[cfg(not(target_family = "wasm"))]
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
```
McPatate, line 142:
```suggestion
#[cfg(not(target_family = "wasm"))]
struct AbortGuard(Option<tokio::task::JoinHandle<()>>);

impl AbortGuard {
    fn abort(self) {
        drop(self);
    }
}

#[cfg(not(target_family = "wasm"))]
impl Drop for AbortOnDrop {
    fn drop(&mut self) {
        if let Some(h) = &self.0 {
            h.abort();
        }
    }
}
```
something like this perhaps? but then again, this is cosmetic feel free to ignore

hf-hub/src/xet.rs
assafvayner: also applied explicit abort in case for upload poller

PR #209, opened 2026-09-07
fix: preserve Xet metadata for streamed downloads

### LINKED ISSUE #208: download_file_stream silently skips Xet after CDN redirects
`download_file_stream()` silently falls back to HTTP for Xet files on native targets. I reproduced this on Windows with 1.0.0; current main has the same code.

For [this model file](https://huggingface.co/black-forest-labs/FLUX.2-small-decoder/resolve/a3efc24f613ef42d9428af62fdbd6f5fd8856c4a/full_encoder_small_decoder.safetensors), the initial HEAD response is a 302 with `X-Xet-Hash` and `X-Linked-Size`. `resolve_xet_hash_and_size()` follows it to the CDN, which returns 200 without the Xet header. The streaming path then skips Xet entirely. A local mock server reproduces the same behavior.

Could this use `head_with_relative_redirects()` and read metadata from the CDN redirect, like the cache path already does?

Found while investigating https://github.com/koharu-rs/koharu/issues/1054.

mayocream:
`download_file_stream()` followed the CDN redirect and lost `X-Xet-Hash`, silently downloading Xet files over HTTP. This reuses the existing relative-redirect helper and reads metadata from the redirect response.

Adds four local regression tests, including two that fail before the fix and pass after it.

Fixes #208.

### DIFF
--- a/hf-hub/src/repository/download.rs
+++ b/hf-hub/src/repository/download.rs
@@ -124,20 +124,20 @@ impl<T: RepoType> HFRepository<T> {
             .hf_client
             .download_url(self.repo_type.url_prefix(), &repo_path, revision, filename)?;
         let headers = self.hf_client.auth_headers();
-        let head_response = retry::retry(self.hf_client.retry_config(), || {
-            self.hf_client.http_client().head(&url).headers(headers.clone()).send()
-        })
-        .await?;
-        let head_response = self
-            .hf_client
-            .check_response(
-                head_response,
-                Some(&repo_path),
-                crate::error::NotFoundContext::Entry {
-                    path: filename.to_string(),
-                },
-            )
-            .await?;
+        let head_response = self.hf_client.head_with_relative_redirects(&url, &headers).await?;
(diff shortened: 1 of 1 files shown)


mayocream:
Plan: reuse the existing metadata HEAD helper, test direct and relative redirects, ordinary files, and missing-file errors locally, and check Xet selection through the public streaming API with a mock token endpoint. Verify the reported public file's headers, formatting, Clippy, and WASM builds.

All four regression tests pass; the two Xet tests failed before the fix. Nightly formatting and `scripts/verify_wasm.sh` pass. The public file's initial 302 carries `X-Xet-Hash`; its CDN response does not.

Clippy is blocked by two existing Windows-only errors in `cache/storage.rs`: the unused `relative` variable at line 77 and a needless borrow at line 90.

assafvayner (approved):
LGTM with some small refactor. this is actually an issue in another code path (get_file_metadata) that will be fixed in a later PR.

PR #215, opened 2026-09-17
Install CI tools with --locked

assafvayner:
The audit job started failing because `cargo install` ignores the workspace lockfile and re-resolves from scratch, so it picked up `cfg-if 1.0.5` hours after publication and the registry proxy rejected it:

```
failed to get successful HTTP response from .../api/v1/crates/cfg-if/1.0.5/download, got 451
{"error":"cfg-if@1.0.5 younger than 3d — too new"}
```

Our own `Cargo.lock` was never involved — it pins `cfg-if 1.0.4` and was resolving fine.

Both `cargo-audit` 0.22.2 and `wasm-bindgen-cli` 0.2.121 ship a `Cargo.lock`, and I checked every package they pin (398 and 261 respectively): none is younger than 3 days, and both pin `cfg-if 1.0.4`. `--locked` therefore builds them from known-good versions rather than whatever landed on crates.io that morning.

`wasm-bindgen-cli` had the same latent failure, so it gets the same treatment.

Also bumps `rustls` 0.23.37 -> 0.23.45 in `Cargo.lock` to clear `RUSTSEC-2026-0285`, which was the remaining audit failure once the install succeeded.

Also bumps every pinned action SHA to its latest release: `actions/checkout` v7.0.1, `actions/cache` v6.1.0, `actions/setup-python` v7.0.0, `Swatinem/rust-cache` v2.9.2, `benchmark-action/github-action-benchmark` v1.22.2, `rust-lang/crates-io-auth-action` v1.0.5, `dtolnay/rust-toolchain` unified on current master, and `hf-workflows` setup-hf-registry-proxies on current main.


McPatate (commented):
Good stuff! I'd take this opportunity to update the actions sha version while we're here.

Also, can you add locked on all workflow files, release and benchmarks?

.github/workflows/rust.yml
```diff
@@ -9,43 +9,49 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+permissions: {}
```
McPatate, line 12: Is this needed?

.github/workflows/benchmarks.yml
```diff
@@ -16,38 +16,42 @@ on:
         required: false
         type: string
 
-permissions:
-  contents: write
-  checks: write
-  pull-requests: write
+
+permissions: {}
```
McPatate, line 20: Same here, are we sure that this is needed?

McPatate (approved):

